PR #1, opened 2020-02-05
feat(root): Initial Infrastructure with Basic Input

mucahit:
### Description
I set up a basic infrastructure to start developing components. The stack is:
- TypeScript & React
- ~Styled Components~ Switched to SCSS
- Jest
- ESLint & Stylelint & Prettier
- Rollup 
- Storybook

### Notes
- ~Please do not mind dummy CSS in the components. They'll be replaced with future PR's.~
- ~`Input` and `FormField` components are a small demo to test the infrastructure.~


src/Input/Input.tsx
```diff
@@ -0,0 +1,103 @@
...
+`;
+const SuffixIcon = styled.span`
+  background: white;
+
+  border-left: 1px solid grey;
+`;
+
+interface InputProps {
+  name: string;
+  type?: string;
```
brkn:
```ts
type InputType =
  | "checkbox"
  | "button"
  | "color"
  | "date"
  | "datetime-local"
  | "email"
  | "file"
  | "hidden"
  | "image"
  | "month"
  | "number"
  | "password"
  | "radio"
  | "range"
  | "reset"
  | "search"
  | "submit"
  | "tel"
  | "text"
  | "time"
  | "url"
  | "week";

interface InputProps {
  name: string;
  type?: InputType;
	// ....
```

src/Form/Field/FormField.tsx
```diff
@@ -0,0 +1,59 @@
...
+    <div className={formFieldClassName}>
+      {hasLabel && (
+        <label id={labelledBy} className={"form-field-label"}>
+          {label}
+        </label>
+      )}
+
+      {children}
+
+      {hasErrorMessage && <ErrorMessage>{errorMessage}</ErrorMessage>}
```
brkn: We have `warningMessage` and `errorMessage` at Wolarf. `helperMessage === warningMessage`?

src/Input/Input.tsx
mucahit: Most of these will be a separate component, probably we only need to expect these `text, email, password, number, url` types. 

I'll make the change when designs are done.

src/Form/Field/FormField.tsx
mucahit: There is a small difference between warning and helper. 

The helper message gives you a small tip about the field for example, "Best usernames are about 5 characters." is a helper message because it gives you a small tip but doesn't restrict you to use username longer than 5 characters. 

In this case, warning/error message would be "Please use a username shorter than 5 characters" and probably you're gonna restrict the user.

mucahit:
Thanks for the review @brkn!

src/Input/Input.tsx
brkn: Okay then 👍 I've discovered we had bunch of `type=string` props, which was not valid. So this is quite useful imo.

src/Form/Field/FormField.tsx
```diff
@@ -0,0 +1,59 @@
...
+    "form-field",
+    {
+      "has-error": hasErrorMessage
+    },
+    className
+  );
+
+  return (
+    <div className={formFieldClassName}>
+      {hasLabel && (
```
burakguneli: label should always be placed on DOM. It's very important for assistive technologies. We can hide it using CSS if necessary.

mucahit: Yeah, but should we show even if it's empty? It's a basic existence check `hasLabel = Boolean(label)`

src/Input/Input.tsx
```diff
@@ -0,0 +1,103 @@
+import React from "react";
+import classNames from "classnames";
+import styled from "styled-components";
+
+const InputContainer = styled.div`
+  display: flex;
+`;
+const PrefixIcon = styled.span`
```
edizcelik: I think naming the icons `left` and `right` is better. `prefix` and `suffix` are for words

src/index.ts
```diff
@@ -0,0 +1,4 @@
+import FormField from "./Form/Field/FormField";
```
edizcelik: I think folder names should be in lowercase, and could be separated with a dash if there is more than one word in the folder name: `form/field/FormField`, `input/Input`.

Should we move `input` inside `form` folder ?

package.json
```diff
@@ -0,0 +1,66 @@
...
+    "babel-loader": "8.0.6",
+    "eslint": "6.6.0",
+    "eslint-config-prettier": "6.5.0",
+    "eslint-plugin-jsx-a11y": "6.2.3",
+    "eslint-plugin-react": "7.16.0",
+    "eslint-plugin-react-hooks": "2.2.0",
+    "fork-ts-checker-webpack-plugin": "4.0.3",
+    "jest": "24.9.0",
+    "prettier": "1.18.2",
+    "react": "16.9.0",
```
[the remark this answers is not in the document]
mucahit: Yes, it's updated with latest prs

PR #2, opened 2020-04-24
feat(root): Implement Button and Spinner components

mucahit:
### Description
- Implemented Button component with basic styles, it supports spinner
- Implemented a basic Spinner component 
- You can style Spinner component with CSS or with props `<Spinner spinnerColor="white"/>`

**If you'd like to see how those components look like, then [check it out here](https://react-ui-toolkit.muco.now.sh/?path=/story/button--button-sizes)**


src/button/Button.tsx
```diff
@@ -0,0 +1,121 @@
...
+  ariaLabel,
+  isDisabled,
+  onMouseOver,
+  onMouseDown,
+  onMouseUp,
+  onFocus,
+  onBlur,
+  tabIndex
+}: ButtonProps) {
+  const shouldInactivate = Boolean(isDisabled || shouldDisplaySpinner);
```
edizcelik, line 51: The name sounds like it causes a side effect to happen. I think `isButtonInactive` or `isButtonDisabled` is a more suited name for this variable. What do you think ?

mucahit: 👍 Going with `isButtonDisabled`. Simple and straight to the point

src/button/Button.tsx
```diff
@@ -0,0 +1,121 @@
...
+      disabled={isButtonDisabled}
+      aria-label={ariaLabel}>
+      {shouldDisplaySpinner ? (
+        <Spinner
+          customClassName={"button-spinner"}
+          aria-label={"Button spinner visible. Button inactivated."}
+        />
+      ) : (
+        children
+      )}
```
jamcry, line 88: Should we hide the button content while displaying spinner? I think we can show spinner next to content (_as in the screenshot_), or make this behavior optional with a prop.
Screen Shot 2020-04-28 at 11 12 36

mucahit: Thanks! I think making it optional would be good

src/button/Button.tsx
```diff
@@ -0,0 +1,121 @@
...
+      type={type}
+      lang={lang}
+      onClick={handleClick}
+      onMouseOver={handleMouseOver}
+      onMouseDown={onMouseDown}
+      onMouseUp={onMouseUp}
+      onFocus={handleFocus}
+      onBlur={onBlur}
+      disabled={isButtonDisabled}
+      aria-label={ariaLabel}>
```
burakguneli, line 80: Let's also add `aria-labelledby` prop in non-required status.

PR #5, opened 2020-04-28
feat(root): Implement List Component

mucahit:
### Description
- List component is done 🎉 
- Currently, it's not supporting `header` & `footer`. Those will be added in the further PR's. 

**To see the implementation [check it out](https://react-ui-toolkit.muco.now.sh/?path=/story/list--has-items)**


src/list/List.tsx
```diff
@@ -0,0 +1,85 @@
...
+  canDisplayEmptyState,
+  placeholders,
+  emptyStateMessage,
+  emptyStateButtonText,
+  onEmptyStateButtonClick
+}: ListProps<Item>) {
+  const listClassName = classNames("list", customClassName);
+
+  return (
+    <div className={listClassName}>
```
edizcelik, line 41: It should be a `ul`

src/list/List.tsx
```diff
@@ -0,0 +1,85 @@
...
+          {Boolean(emptyStateButtonText && onEmptyStateButtonClick) && (
+            <Button
+              customClassName={"list-empty-state-button"}
+              testid={"list.empty-state-button"}
+              onClick={onEmptyStateButtonClick}>
+              {emptyStateButtonText}
+            </Button>
+          )}
+        </div>
+      )}
```
edizcelik, line 79: I really don't think these should be the responsibility of a List component. They unnecessarily increase the props of the component and complicates the concerns. The only concern of the List component should be to display `ul` and its items. Pending and empty states should be handled separately.

src/list/List.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+/* eslint-disable complexity */
+function List<Item extends any>({
+  items,
+  children,
+  customClassName,
+  testid,
+  listItemKeyGenerator,
+  placeholderProps,
```
bilgeyucel, line 29:
```suggestion
  pendingStateProps,
```
Wouldn't it be better if it was named `pendingStateProps`? I cannot come up with a different use case and `emptyStateProps` is a placeholder, too.

PR #8, opened 2020-07-06
chore: Switch to CSS Variables

mucahit:
### Firstly Merge #9

### DIFF
--- a/src/ui/reference/_colors.scss
+++ b/src/ui/reference/_colors.scss
@@ -1,11 +1,34 @@
 /* stylelint-disable color-no-hex */
-$black: black;
-$dark-grey: #484848;
-$grey: #757575;
-$light-grey: #EBEBEB;
 
-$red: #A43030;
+:root {
+  // Button
+  --button-bg: black;
 
-$default-box-shadow: 0 0 5px $light-grey;
+  // Dropdown
+  --dropdown-text-color: black;
+  --dropdown-header-button-bg: #ebebeb;
+  --dropdown-selected-item-bg: #ebebeb;
+
+  // Checkbox
+  --checkbox-border-color: black;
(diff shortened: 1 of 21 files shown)


src/ui/reference/_colors.scss
```diff
@@ -1,11 +1,34 @@
...
-$red: #A43030;
+:root {
+  // Button
+  --button-bg: black;
 
-$default-box-shadow: 0 0 5px $light-grey;
+  // Dropdown
+  --dropdown-text-color: black;
+  --dropdown-header-button-bg: #ebebeb;
+  --dropdown-selected-item-bg: #ebebeb;
```
edizcelik, line 10: I think this file will increasingly get larger as we add more components to the UI-kit. Can we define the variables in respective SCSS files for the components? For example, dropdown variables can be defined inside `_dropdown.scss`.

src/ui/reference/_colors.scss
```diff
@@ -1,11 +1,34 @@
...
 
-$red: #A43030;
+:root {
+  // Button
+  --button-bg: black;
 
-$default-box-shadow: 0 0 5px $light-grey;
+  // Dropdown
+  --dropdown-text-color: black;
+  --dropdown-header-button-bg: #ebebeb;
```
bilgeyucel, line 9: I was thinking maybe we can still have variables for colors.

```
$light-grey: #ebebeb;
```
and then 
```suggestion
  --dropdown-header-button-bg: $light-grey;
```

src/ui/reference/_colors.scss
mucahit: Filed an issue https://github.com/Hipo/react-ui-toolkit/issues/11

PR #9, opened 2020-07-06
feat(tab): Add Tab Component

mucahit:
### Description
- Basic implementation of Tab component

### DIFF
--- a/src/tab/Tab.tsx
+++ b/src/tab/Tab.tsx
@@ -3,7 +3,7 @@ import "./_tab.scss";
 import React, {useState} from "react";
 import classNames from "classnames";
 
-import Button from "../button/Button";
+import TabHeaderItem from "./header/item/TabHeaderItem";
 
 export type TabItem = {
   id: string;
@@ -13,26 +13,26 @@ export type TabItem = {
 
 interface TabProps {
   items: TabItem[];
+  initialActiveTabIndex?: number;
   children: React.ReactNode[];
   customClassName?: string;
 }
 
-function Tab({items, children, customClassName}: TabProps) {
-  const [activeTabIndex, setActiveTabIndex] = useState(0);
+function Tab({items, initialActiveTabIndex = 0, children, customClassName}: TabProps) {
+  const [activeTabIndex, setActiveTabIndex] = useState(initialActiveTabIndex);
   const tabClassName = classNames("tab", customClassName);
 
   return (
     <div className={tabClassName}>
       <header className={"tab-header"}>
         {items.map((item, index) => (
-          <Button
+          <TabHeaderItem
             key={item.id}
-            customClassName={"tab-header-item"}
-            onClick={handleChangeActiveTab(index)}>
-            {item.icon && <span className={"tab-header-item-icon"}>{item.icon}</span>}
-
+            onClick={handleChangeActiveTab}
+            icon={item.icon}
+            index={index}>
             {item.children}
-          </Button>
+          </TabHeaderItem>
         ))}
       </header>
 
@@ -41,7 +41,7 @@ function Tab({items, children, customClassName}: TabProps) {
   );
 
   function handleChangeActiveTab(index: number) {
-    return () => setActiveTabIndex(index);
+    setActiveTabIndex(index);
   }
 }
 
--- a/src/tab/Tab.tsx
+++ b/src/tab/Tab.tsx
@@ -1,0 +1,48 @@
+import "./_tab.scss";
+
+import React, {useState} from "react";
+import classNames from "classnames";
+
+import Button from "../button/Button";
+
+export type TabItem = {
+  id: string;
+  children: React.ReactNode;
+  icon?: React.ReactNode;
+};
+
+interface TabProps {
+  items: TabItem[];
+  children: React.ReactNode[];
+  customClassName?: string;
+}
+
+function Tab({items, children, customClassName}: TabProps) {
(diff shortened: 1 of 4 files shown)


src/tab/Tab.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const tabClassName = classNames("tab", customClassName);
+
+  return (
+    <div className={tabClassName}>
+      <header className={"tab-header"}>
+        {items.map((item, index) => (
+          <Button
+            key={item.id}
+            customClassName={"tab-header-item"}
+            onClick={handleChangeActiveTab(index)}>
```
jamcry, line 31: This looks odd to me, at first glance I thought this was a bug (function called inside `onClick` prop) 😄  I know it is against our `eslint` rules, but I think in this case inline arrow functions can be more readable. And it should be same as current one in performance aspect (which is the main concern of `no-bind` rule)
Or, of course, we can split tab button to its own component and declare event handler in there. What do you think?

src/tab/Tab.tsx
```diff
@@ -0,0 +1,48 @@
...
+};
+
+interface TabProps {
+  items: TabItem[];
+  children: React.ReactNode[];
+  customClassName?: string;
+}
+
+function Tab({items, children, customClassName}: TabProps) {
+  const [activeTabIndex, setActiveTabIndex] = useState(0);
```
edizcelik, line 21: I think it might be useful if we could pass down a prop to control the initial value of the  `activeTabIndex` state. It can default to `0`

src/tab/Tab.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const tabClassName = classNames("tab", customClassName);
+
+  return (
+    <div className={tabClassName}>
+      <header className={"tab-header"}>
+        {items.map((item, index) => (
+          <Button
+            key={item.id}
```
edizcelik, line 29: We should probably be adding `testids` otherwise we will get complaints from the QA team :)

src/tab/Tab.tsx
```diff
@@ -0,0 +1,48 @@
...
+            customClassName={"tab-header-item"}
+            onClick={handleChangeActiveTab(index)}>
+            {item.icon && <span className={"tab-header-item-icon"}>{item.icon}</span>}
+
+            {item.children}
+          </Button>
+        ))}
+      </header>
+
+      <div className={"tab-body"}>{children[activeTabIndex]}</div>
```
edizcelik, line 39:
```suggestion
      {children[activeTabIndex]}
```

Let's provide a `TabBody` component:
```suggestion
      function TabBody({children, className}: {children: React.ReactNode; className?: string;}) {
        return <div className={classNames("tab-body", className)}>{children}</div>
      }
```

src/tab/_tab.scss
```diff
@@ -0,0 +1,12 @@
...
+  .tab-header {
+    display: flex;
+  }
+
+  .tab-header-item {
+    --button-bg: transparent;
+    --button-width: auto;
+
+    color: black;
+  }
```
edizcelik, line 11: Let's not nest the selectors if it isn't necessary. This will make overriding easier

stories/6-Tab.stories.tsx
```diff
@@ -0,0 +1,21 @@
...
+  },
+  {
+    id: "following",
+    children: "Following"
+  }
+];
+
+storiesOf("Tab", module).add("Tab", () => (
+  <Fragment>
+    <Tab items={tabItems}>{[<div>{"Home"}</div>, <div>{"Following"}</div>]}</Tab>
```
edizcelik, line 19: Here we would use `TabBody`:

```suggestion
    <Tab items={tabItems}>
      {[
        <TabBody>{"Home"}</TabBody>,
        <TabBody className={"following-tab"}>{"Following"}</TabBody>
      ]}
    </Tab>
```

src/tab/Tab.tsx
edizcelik: I think it would be the best to define `TabHeaderItem` component

stories/6-Tab.stories.tsx
bilgeyucel: Wouldn't it be better if we handle wrapping with `<div className={classNames("tab-body", className)}>` inside `Tab` as @mucahit did in the first place? Then we can pass the child component directly and won't have to import `TabBody` every time we use `Tab`.

src/tab/Tab.tsx
mucahit: Thanks! I'll keep this as it is for now 👍

src/tab/header/item/TabHeaderItem.tsx
```diff
@@ -0,0 +1,28 @@
...
+  index: number;
+  onClick: (index: number) => void;
+};
+
+function TabHeaderItem({children, icon, index, onClick}: TabHeaderItemProps) {
+  return (
+    <Button
+      customClassName={"tab-header-item"}
+      onClick={handleClick}
+      testid={`tab-header-item-${index}`}>
```
edizcelik, line 16: Let's pass `testid` as props to this component. Similarly, I think `Tab` component should also get a `testid` and the testid for each TabHeaderItem should be `${testidForTab}.TabHeaderItem-${index}`

mucahit: Filed an issue https://github.com/Hipo/react-ui-toolkit/issues/12

PR #25, opened 2020-09-16
fix(dropdown): Improvements on styles and handle menu visibility via CSS

edizcelik:
### Description

- `document.hidden` is used on dropdown focus and blur handlers to handle changes on screen visibility correctly
- Instead of toggling the whole dropdown menu from DOM, we now use CSS to handle its visibility
- Effect that listens `selectOption` to focus on it is removed and now we only use `focusedOption` to handle scroll computations
- Direct DOM access to the focused list item element is removed to use React Ref API.
- Nesting of `dropdown-header-button` CSS block is removed

### DIFF
--- a/src/dropdown/Dropdown.tsx
+++ b/src/dropdown/Dropdown.tsx
@@ -1,8 +1,6 @@
 /* eslint-disable max-lines */
 import CaretDownIcon from "../ui/icons/caret-down.svg";
 
-import "./_dropdown.scss";
-
 import React, {Fragment, useState, useRef, useEffect} from "react";
 import classNames from "classnames";
 
@@ -27,6 +25,8 @@ import {
 import Spinner from "../spinner/Spinner";
 import {SINGLE_ALPHANUMERIC_CHARACTER_REGEX} from "../core/utils/stringConstants";
 import useOnClickOutside from "../core/utils/hooks/onClickOutside";
+
+import "./_dropdown.scss";
 
 export type MenuVisibilityChangeHandlerTypeArgument = "open" | "closed";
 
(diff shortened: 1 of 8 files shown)


src/dropdown/Dropdown.tsx
```diff
@@ -28,6 +26,8 @@ import Spinner from "../spinner/Spinner";
 import {SINGLE_ALPHANUMERIC_CHARACTER_REGEX} from "../core/utils/stringConstants";
 import useOnClickOutside from "../core/utils/hooks/onClickOutside";
 
+import "./_dropdown.scss";
```
jamcry, line 29: I believe we should add a comment here explaining why it is imported here to prevent ambiguity

src/dropdown/list/_dropdown-list.scss
```diff
@@ -2,15 +2,53 @@
...
+
+  &.right {
+    top: 0;
+    right: 0;
+
+    transform: translate(100%, -50%);
+  }
+}
+
+.dropdown-list__is-visible {
```
jamcry, line 49:
```suggestion
.dropdown-list--is-visible {
```
Shouldn't we separate `is-visible` with `--` because it represents the state of the element (i.e. a modifier)?

bilgeyucel (approved):

mucahit (approved):

burakguneli (approved):

src/dropdown/list/_dropdown-list.scss
edizcelik: Yep, thanks for catching this!

jamcry (approved):

PR #26, opened 2020-09-25
fix(dropdown): Set dropdown background color using css variable

jamcry:


mucahit (approved):
Thanks!

src/dropdown/_dropdown.scss
```diff
@@ -43,7 +43,7 @@ $dropdown-header-horizontal-padding: 8px;
   width: 100%;
   height: var(--dropdown-header-button-height);
 
-  background-color: white;
+  background-color: var(--dropdown-header-button-bg);
   border: var(--dropdown-header-button-border-width) solid var(--default-border-color);
```
edizcelik, line 47: Should we have a variable for border-style as well ? 😄

src/dropdown/_dropdown.scss
```diff
@@ -43,7 +43,7 @@ $dropdown-header-horizontal-padding: 8px;
   width: 100%;
   height: var(--dropdown-header-button-height);
```
edizcelik, line 44: I think we can define these via `--button-width` and `--button-height`, can we ?

mucahit: Agreed. Let's do it in another PR

src/dropdown/_dropdown.scss
mucahit: Yeah maybe we can do something like this

```css
  --dropdown-header-button-border: var(--dropdown-header-button-border-width) solid var(--default-border-color)
```

```suggestion
  border: var(--dropdown-header-button-border);
```

bilgeyucel (approved):

PR #41, opened 2020-11-03
fix(tab): Add `active` className

mucahit:
### Description
- Added a className to active `TabHeaderItem`

### DIFF
--- a/src/tab/header/item/TabHeaderItem.tsx
+++ b/src/tab/header/item/TabHeaderItem.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import classNames from "classnames";
 
 import {TabItem} from "../../Tab";
 import Button from "../../../button/Button";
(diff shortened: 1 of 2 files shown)


src/tab/header/item/TabHeaderItem.tsx
```diff
@@ -1,17 +1,19 @@
...
   onClick: (index: number) => void;
+  isActive: boolean;
 };
 
-function TabHeaderItem({children, icon, index, onClick}: TabHeaderItemProps) {
+function TabHeaderItem({children, icon, index, onClick, isActive}: TabHeaderItemProps) {
   return (
     <Button
-      customClassName={"tab-header-item"}
+      customClassName={classNames("tab-header-item", {active: isActive})}
```
burakguneli, line 16: Shouldn't you wrap `active` with quotation marks? Like: `{"active": isActive}`

burakguneli (changes requested):

jamcry (approved):

src/tab/header/item/TabHeaderItem.tsx
jamcry: I think our prettier rules automatically removes `"`s from keys if they don't include special characters (such as `-`)

edizcelik: Agreed with other comments but let's use BEM convention here
```suggestion
      customClassName={classNames("tab-header-item", {"tab-header-item--is-active": isActive})}
```

mucahit: Done 👍

bilgeyucel (approved):

PR #48, opened 2020-12-30
chore: General Improvements

### LINKED ISSUE #28: List: Should allow passing `role` attribute
When I try to set the role as `navigation` for the List component, I'm faced with the following error:

```
Type '{ children: (tab: any, testid: string) => Element; role: string; "data-testid": string; items: any[]; customClassName: string; }' is not assignable to type 'IntrinsicAttributes & ListProps<any>'.
  Property 'role' does not exist on type 'IntrinsicAttributes & ListProps<any>'.ts(2322)
```

### LINKED ISSUE #35: Add `testid`s to RadioGroup items

### LINKED ISSUE #40: FormField: Allow testid as a prop
FormField should get a testid prop and provide `FormFieldMessageRow` with necessary testids.

### LINKED ISSUE #42: Checkbox: onSelect should be able to return context
```
export interface CheckboxInputItem {
    testid?: string;
    id: string;
    content: React.ReactNode;
    inputProps: {
        htmlFor: string;
        value: string;
        name: string;
    };
}
export interface CheckboxInputProps {
    item: CheckboxInputItem;
    onSelect: (name: string, item: CheckboxInputItem) => void;
    isSelected: boolean;
    isDisabled?: boolean;
    customClassName?: string;
}
```

we should add one parameter to `onSelect` so that instead of using only one string field we can use the whole context of the option

### LINKED ISSUE #43: Let's omit <disabled> prop that comes from React.InputHTMLAttributes<HTMLInputElement>
As we have a separate `isDisabled` prop, we should omit `disabled` prop that comes from React.InputHTMLAttributes<HTMLInputElement> so that we disallow using `disabled` as prop to Input component. Or remove `isDisabled` in favour of using `disabled` prop.
https://github.com/Hipo/react-ui-toolkit/blob/master/src/form/input/Input.tsx#L8

I think we should clean the typing here so that there are no duplications. For example, we should remove `onFocus` and `onBlur` as they are already available from `React.InputHTMLAttributes<HTMLInputElement>` in the form that we want. Also, `onChange` is required for our Input component however we should allow it to be optional which is how it is defined within `React.InputHTMLAttributes<HTMLInputElement>`. For example, some inputs might always be disabled so there is no need to define a `onChange` for them.

```
export type InputProps = Omit<React.InputHTMLAttributes<HTMLInputElement>, "disabled" | "name">
 & {
  name: string;
  testid?: string;
  leftIcon?: React.ReactNode;
  rightIcon?: React.ReactNode;
  isDisabled?: boolean;
  hasError?: boolean;
  customClassName?: string;
  inputContainerRef?: React.RefObject<HTMLDivElement>;
}
```

mucahit:
### Related Issues
- Resolves #28
- Resolves #35 
- Resolves #40
- Resolves #42 
- Resolves #43

### DIFF
--- a/src/form/field/FormField.tsx
+++ b/src/form/field/FormField.tsx
@@ -12,6 +12,7 @@ export interface FormFieldProps {
   customClassName?: string;
   helperMessages?: Array<string>;
   errorMessages?: Array<string>;
+  testid?: string;
 }
 
 function FormField(props: FormFieldProps) {
@@ -22,7 +23,8 @@ function FormField(props: FormFieldProps) {
     customClassName,
     children,
     errorMessages,
-    helperMessages
+    helperMessages,
+    testid
   } = props;
   const hasErrorMessage = Boolean(errorMessages?.length);
   const hasHelperMessage = Boolean(helperMessages?.length);
@@ -35,7 +37,7 @@ function FormField(props: FormFieldProps) {
   );
 
   return (
-    <div className={formFieldClassName}>
+    <div className={formFieldClassName} data-testid={testid}>
       <label id={labelledBy} htmlFor={labelFor} className={"form-field-label"}>
         {Boolean(label) && <span className={"form-field-label-text"}>{label}</span>}
 
@@ -48,6 +50,7 @@ function FormField(props: FormFieldProps) {
             key={`error.${message}`}
             type={"error"}
             message={message}
+            testid={`${testid}.form-field-message-row.error`}
           />
         ))}
 
@@ -58,6 +61,7 @@ function FormField(props: FormFieldProps) {
             key={`helper.${message}`}
             type={"helper"}
             message={message}
+            testid={`${testid}.form-field-message-row.helper`}
           />
         ))}
     </div>
--- a/src/form/input/checkbox/Checkbox.tsx
+++ b/src/form/input/checkbox/Checkbox.tsx
@@ -6,7 +6,6 @@ import React from "react";
 import classNames from "classnames";
 
 export interface CheckboxInputItem {
-  testid?: string;
   id: string;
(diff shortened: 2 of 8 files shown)


src/form/input/checkbox/Checkbox.tsx
```diff
@@ -6,19 +6,23 @@ import React from "react";
...
 export interface CheckboxInputItem {
-  testid?: string;
   id: string;
   content: React.ReactNode;
   inputProps: {
     htmlFor: string;
     value: string;
     name: string;
   };
+  context?: any;
```
edizcelik, line 16: We can give the type for context as a generic which defaults to `any` if not provided

edizcelik (approved):

src/form/input/checkbox/Checkbox.tsx
jamcry: I think that can be useful in some cases 👍 Also, same for `Radio` component

src/form/field/FormField.tsx
```diff
@@ -48,6 +50,7 @@ function FormField(props: FormFieldProps) {
             key={`error.${message}`}
             type={"error"}
             message={message}
+            testid={`${testid}.form-field-message-row.error`}
```
jamcry, line 53: Should we set testid to some default value (in line 27)?
Otherwise if testid is not given, I think this will result: `undefined.form-field-message...`

edizcelik: Good point but it doesn't hurt anything for now so I can keep it to read undefined for now.

edizcelik:
I am merging this now. A related issue is opened: https://github.com/Hipo/react-ui-toolkit/issues/49

Issue #49, opened 2021-01-04
Add a generic type for context for Checkbox and Radio components


edizcelik:
See comments here: https://github.com/Hipo/react-ui-toolkit/pull/48#discussion_r550235892

Issue #36, opened 2020-10-13
Cannot import `RadioInputItem` type


edizcelik:
When I try to use `RadioInputItem`, it imports it like the following:

```
import {RadioInputItem} from "@hipo/react-ui-toolkit/dist/form/input/radio/Radio";
```

However, this crushes the app with the following message:

```
Error: Cannot find module '@hipo/react-ui-toolkit/dist/form/input/radio/Radio'
at webpackMissingModule (http://dylan.dev.mastermind.co:3000/product-flow.bundle.js:3612:45)
at Module.<anonymous> (http://dylan.dev.mastermind.co:3000/product-flow.bundle.js:3612:170)
at Module../src/product/add/live-event/form/AddProductLiveEventForm.tsx (http://dylan.dev.mastermind.co:3000/product-flow.bundle.js:3792:30)
at __webpack_require__ (http://dylan.dev.mastermind.co:3000/main.bundle.js:854:30)
at fn (http://dylan.dev.mastermind.co:3000/main.bundle.js:151:20)
at Module.<anonymous> (http://dylan.dev.mastermind.co:3000/product-flow.bundle.js:3453:87)
at Module../src/product/add/live-event/AddProductLiveEventPage.tsx (http://dylan.dev.mastermind.co:3000/product-flow.bundle.js:3509:30)
at __webpack_require__ (http://dylan.dev.mastermind.co:3000/main.bundle.js:854:30)
at fn (http://dylan.dev.mastermind.co:3000/main.bundle.js:151:20)
at Module.<anonymous> (http://dylan.dev.mastermind.co:3000/product-flow.bundle.js:5197:98)
```

Issue #27, opened 2020-10-05
Reduce specificity within dropdown styles


edizcelik:
~~This selector is too specific and I think we should be able to reduce its specificity~~

```
.dropdown-container button.dropdown-header-button
```

~~Also, the `height` property is set via the height property in this block and I think we should remove that in favour of using `--button-height` variable~~  This was a mistake on my side on the project.

mucahit:
Is this still valid? @edizcelik

edizcelik:
What I said about `dropdown-header-button` is invalid. However, I think we can still improve the dropdown styles a bit:
`has-header-box-shadow` and `disabled` is currently nested and we can remove the nesting there by using BEM naming practices.

Issue #12, opened 2020-07-16
Add `testid` to Tab Component


mucahit:
See the related comment https://github.com/Hipo/react-ui-toolkit/pull/9#pullrequestreview-447126215

PR #51, opened 2021-01-05
feat: Makes changes to follow BEM convention when defining classnames & other improvements

edizcelik:
### Related Issues

- Resolves https://github.com/Hipo/react-ui-toolkit/issues/11
- Resolves https://github.com/Hipo/react-ui-toolkit/issues/49
- Resolves https://github.com/Hipo/react-ui-toolkit/issues/37
- Resolves https://github.com/Hipo/react-ui-toolkit/issues/36
- Resolves https://github.com/Hipo/react-ui-toolkit/issues/27
- Resolves https://github.com/Hipo/react-ui-toolkit/issues/12


src/dropdown/list/_dropdown-list.scss
```diff
@@ -47,7 +47,3 @@
   visibility: visible;
   opacity: 1;
 }
-
-.dropdown-empty-message {
-  padding: 0 15px;
-}
```
jamcry, line 53: This rule is removed, but it wasn't added with the new name (`dropdown-list__empty-message`)
Is this rule not necessary anymore?

src/tab/header/item/_tab-header-item.scss
```diff
@@ -0,0 +1,25 @@
+.tab-header-item {
+  --active-color: 50, 50, 211;
```
jamcry, line 2: Should we make this variable name more specific (maybe `tab-header-active-color`) ? It sounds too generic so it may conflict with other variable names

stories/utils/StoryFragment.tsx
```diff
@@ -0,0 +1,15 @@
+import "../../src/ui/reference/_colors.scss";
```
jamcry, line 1: I couldn't figure out the purpose of this component 🤔

mucahit (approved):
Thanks! 💯

src/dropdown/list/_dropdown-list.scss
edizcelik: Nope there is no need to give it a padding anymore

stories/utils/StoryFragment.tsx
edizcelik: Otherwise default css variables within _colors or _measurement were not showing up on Storybook. So its a bit of a hack

PR #53, opened 2021-02-09
feat(typeahead/select): Remove tags with backspace key

### LINKED ISSUE #47: TypeaheadSelect: Remove tags with Backspace key
When there is no value within the input, hitting backspace key should pop selected options.

Anlerkan:
### Related Issues

- Resolves #47

### DIFF
--- a/src/select/typeahead/TypeaheadSelect.tsx
+++ b/src/select/typeahead/TypeaheadSelect.tsx
@@ -192,14 +192,13 @@ function TypeaheadSelect({
   function handleKeyDown(event: React.KeyboardEvent<HTMLInputElement>) {
     const {key} = event;
 
-    if (key === KEYBOARD_EVENT_KEY.ESCAPE) {
-      if (!computedDropdownOptions.length) {
-        event.stopPropagation();
-
-        if (onTagRemove) {
-          onTagRemove(selectedOptions[selectedOptions.length - 1]);
-        }
-      }
+    if (
+      key === KEYBOARD_EVENT_KEY.ESCAPE &&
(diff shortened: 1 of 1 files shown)


src/select/typeahead/TypeaheadSelect.tsx
```diff
@@ -187,6 +189,19 @@ function TypeaheadSelect({
     }
   }
 
+  function handleKeyDown(event: React.KeyboardEvent<HTMLInputElement>) {
+    const {key} = event;
+
+    if (
+      key === KEYBOARD_EVENT_KEY.ESCAPE &&
```
jamcry, line 196:
```suggestion
      key === KEYBOARD_EVENT_KEY.BACKSPACE &&
```
I think there is a mistake

src/select/typeahead/TypeaheadSelect.tsx
```diff
@@ -187,6 +189,19 @@ function TypeaheadSelect({
...
+      !computedDropdownOptions.length &&
```
jamcry, line 197: I assume this line aims to prevent removing tags in case user had entered a part of keyword and trying to remove a character.
But I think this won't work if user entered a keyword, but there are no matching options (`!computedDropdownOptions.length` is `true`). In this case if user wants to remove the last character of the search keyword, he will end up removing the tag instead.

It may be better to check the input value state directly:
(Also I believe we should check if there are any visible tags, otherwise we don't need to run this block)
```suggestion
      !typeaheadInputValue &&
      tags.length
```

jamcry (changes requested):

PR #54, opened 2021-02-09
feat(typeahead/select): Add a class when there are selectedOptions

### LINKED ISSUE #44: TypeaheadSelect: Add a class when there are selectedOptions
Let's add a class to the container so that it makes it easier to style the component accordingly when there are selectedOptions: `typeahead-select--has-selected-options`

Anlerkan:
### Related Issues

- Resolves #44

### DIFF
--- a/src/select/typeahead/TypeaheadSelect.tsx
+++ b/src/select/typeahead/TypeaheadSelect.tsx
@@ -53,28 +53,27 @@ function TypeaheadSelect({
   canOpenDropdownMenu = true,
   areOptionsFetching
 }: TypeaheadSelectProps) {
+  const typeaheadInputRef = useRef<HTMLDivElement | null>(null);
   const [isMenuOpen, setMenuVisibility] = useState(false);
   const [computedDropdownOptions, setComputedDropdownOptions] = useState(dropdownOptions);
   const [shouldResetTypeaheadValue, setShouldResetTypeaheadValue] = useState(false);
   const [shouldFocusOnInput, setShouldFocusOnInput] = useState(false);
-  const typeaheadSelectClassName = classNames(
-    "typeahead-select__dropdown",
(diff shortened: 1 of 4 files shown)


src/select/typeahead/TypeaheadSelect.tsx
```diff
@@ -71,7 +71,8 @@ function TypeaheadSelect({
     Boolean(selectedOptionLimit && selectedOptions.length >= selectedOptionLimit - 1);
   const typeaheadClassName = classNames("typeahead-select__input", {
     "typeahead-select__input--is-dropdown-menu-open": isMenuOpen,
-    "typeahead-select__input--can-select-multiple": canSelectMultiple
+    "typeahead-select__input--can-select-multiple": canSelectMultiple,
+    "typeahead-select--has-selected-options": Boolean(selectedOptions.length)
```
jamcry, line 75:
```suggestion
    "typeahead-select__input--can-select-multiple": canSelectMultiple,
    "typeahead-select__input--has-selected-options": Boolean(selectedOptions.length)
```
I think it should start with `typeahead-select__input...` because the classname is passed to the input

jamcry: I am sorry, I think I wasn't aware of the intention of the change the first time.
I guess `typeahead-select--has-selected-options` class name was correct, but it should be added to `typeaheadSelectClassName`, not to the input class name, this way it will be applied to the top container, not just the input. This way it will give more control for the styling

```tsx
  const typeaheadSelectClassName = classNames(
    "typeahead-select__dropdown",
    "typeahead-select--has-selected-options",
    customClassName
  );
```

edizcelik: Yea this should've been added to the container

PR #55, opened 2021-02-09
feat(components) : Add Avatar component

### LINKED ISSUE #21: Add Avatar Component
Gets `src, alt, customClassName` properties and gives rounded `img` element with a fallback placeholder image

Anlerkan:
### Description

- Basic implementation of Avatar component

### Related Issues 

- Resolves #21 

### Notes

image


src/avatar/Avatar.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+export interface AvatarProps {
+  alt: string;
+  size?: {
+    width: string;
+    height: string;
+  };
+  placeholderSrc: string;
+  src?: string | null;
+  customClassName?: string;
```
jamcry, line 14:
```suggestion
  placeholderSrc: string;
  size?: {
    width: string;
    height: string;
  };
  src?: string | null;
  customClassName?: string;
```
Let's group optional props at the bottom to follow our conventions

src/avatar/Avatar.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <img
+      src={shouldDisplayPlaceholder ? placeholderSrc : src || placeholderSrc}
+      alt={alt}
+      style={{...size}}
+      className={classNames("avatar", customClassName)}
+      onLoad={handleHidePlaceholder}
+    />
+  );
+
+  function handleHidePlaceholder() {
```
jamcry, line 30: I think `handleImageLoad()` may be a more descriptive name for this function, because here we are handling the load event of the image
```suggestion
  function handleImageLoad() {
```

Anlerkan: Thanks for your suggestions. I applied this changes [@this commit ](https://github.com/Hipo/react-ui-toolkit/pull/55/commits/d59c58b6d37a862c9972aa296516d09122df7e3f) :pray:

jamcry (approved):
thanks 👏

src/avatar/Avatar.tsx
```diff
@@ -0,0 +1,35 @@
+import "./_avatar.scss";
+
+import React, {useState} from "react";
+import classNames from "classnames";
+
+export interface AvatarProps {
+  alt: string;
+  placeholderSrc: string;
```
edizcelik, line 8: We might not a placeholderSrc available. I think this should be optional

src/avatar/Avatar.tsx
```diff
@@ -0,0 +1,35 @@
...
+  size?: {
+    width: string;
+    height: string;
+  };
+  src?: string | null;
+  customClassName?: string;
+}
+
+function Avatar({alt, size, src, customClassName, placeholderSrc}: AvatarProps) {
+  const [shouldDisplayPlaceholder, setPlaceholderVisibility] = useState(true);
```
edizcelik, line 18: I don't understand the need for this state. After making the `placeholderSrc` optional, we can just check if the value is truthy and render `img` with `placeholderSrc` if `src` is falsy. If both of them are falsy, then we shouldn't render any `img` tag I think.

Actually, I am not really sure why we are adding a separate `placeholderSrc` prop. As a user of this component, I could just provide `src` appropriately if I need to render a placeholder image.

src/avatar/Avatar.tsx
```diff
@@ -0,0 +1,35 @@
...
+import React, {useState} from "react";
+import classNames from "classnames";
+
+export interface AvatarProps {
+  alt: string;
+  placeholderSrc: string;
+  size?: {
+    width: string;
+    height: string;
+  };
```
edizcelik, line 12: We always have the avatar as a circle so we can just get a `size` 
```suggestion
  size?: number;
```

yasincaliskan (approved):

src/avatar/Avatar.tsx
```diff
@@ -0,0 +1,24 @@
...
+  size?: number;
+  customClassName?: string;
+}
+
+function Avatar({alt, size, src, customClassName}: AvatarProps) {
+  return (
+    <img
+      src={src}
+      alt={alt}
+      style={{width: `${size}px`, height: `${size}px`}}
```
edizcelik, line 18: What happens if `size` is undefined ? Does it work as expected ? I think we should check if the size is truthy before assigning the style object

bilgeyucel (approved):

edizcelik (approved):

PR #58, opened 2021-03-31
fix(typeahead-select): Improves the handling of controlled keyword state

edizcelik:
### Changes

**`TypeaheadInput` component**

- new `initialValue` prop added and it defaults to empty string.
- when `value` prop is present and it has a string value, this means that TypeaheadInput's value is held as a state in the parent and should be controlled by the parent as well. So whenever value changes and it has a string value then we update the TypeaheadInput's state. TypeaheadInput still controls its own state as it has to deal with debouncing.
- Removes `shouldResetValue` prop.

**`TypeaheadSelect` component**

- `typeaheadProps` prop is updated to have the following type: `Pick<TypeaheadInputProps, "id" | "placeholder" | "name" | "onFocus">`. Note how you cannot pass `value` under this object any longer.
- new `initialKeyword` prop added and it defaults to empty string.
- TypeaheadSelect now resets the input value by the default when an option is selected.
- new `controlledKeyword` prop is added. When this prop is present and it has a non-undefined value it means that the keyword is held as a state in the parent and therefore should be controlled by the parent. When this prop is present, `TypeaheadSelect` never updates its `keyword` state (a good example of inversion of control).

**Stories**

- Updated Typeahead stories so that each typeahead example holds its own state and not affected by the changes on other typeaheads within the page.
- Added an example story about how one can use `controlledKeyword`.

### DIFF
--- a/src/form/input/typeahead/TypeaheadInput.tsx
+++ b/src/form/input/typeahead/TypeaheadInput.tsx
@@ -5,15 +5,16 @@ import Input from "../Input";
 import useDebounce from "../../../core/utils/hooks/debounce";
 
 export interface TypeaheadInputProps {
+  onQueryChange: (value: string) => void;
+  value?: string;
   testid?: string;
   customClassName?: string;
   id?: string;
   name: string;
   isDisabled?: boolean;
-  value?: string;
+  initialValue?: string;
   placeholder: string;
   queryChangeDebounceTimeout?: number;
-  onQueryChange: (value: string) => void;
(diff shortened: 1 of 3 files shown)


mucahit (approved):
Thanks @edizcelik!

Anlerkan (approved):

bilgeyucel (approved):
💯

yasincaliskan (approved):

jamcry (approved):

PR #60, opened 2021-04-01
feat(dropdown): Allow disabled options

edizcelik:


yasincaliskan (approved):

Anlerkan (approved):

src/dropdown/Dropdown.tsx
```diff
@@ -235,7 +235,7 @@ function Dropdown<OptionIdShape extends string>({
     option: Parameters<DropdownOptionSelectHandler>[0],
     event?: React.SyntheticEvent<HTMLLIElement>
   ) {
-    if (!isDisabled && option) {
+    if (!isDisabled && option && !option.isDisabled) {
```
jamcry, line 238: It looks like we prevent clicks on disabled items, should we still check here? (and line 251)

jamcry (approved):

src/dropdown/Dropdown.tsx
edizcelik: You are right. This function should not run at all if it is a disabled option. I removed these checks here and added them for the keydown to make sure we don't call this function when Enter is hit on disabled option. Ideally, when navigating via keyboard, the focus should jump over the disabled option. We can add that functionality later on.

Issue #56, opened 2021-02-16
Input: Color type is not predefined for using color-picker


yasincaliskan:
We need the type of "color" for using the [color-picker](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/color). But isn't defined for now.

PR #63, opened 2021-04-06
fix(input): Add type prop for color picker

yasincaliskan:
### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/56

### DIFF
--- a/src/form/input/Input.tsx
+++ b/src/form/input/Input.tsx
@@ -3,34 +3,36 @@ import "./_input.scss";
 import React from "react";
 import classNames from "classnames";
 
+type InputTypes =
+  | "checkbox"
+  | "button"
+  | "color"
+  | "date"
+  | "datetime-local"
+  | "email"
+  | "file"
+  | "hidden"
+  | "image"
+  | "month"
+  | "number"
+  | "password"
+  | "radio"
+  | "range"
+  | "reset"
+  | "search"
+  | "submit"
+  | "tel"
+  | "text"
+  | "time"
+  | "url"
+  | "week";
+
 export type InputProps = Omit<
   React.InputHTMLAttributes<HTMLInputElement>,
   "disabled" | "name" | "className"
 > & {
   name: string;
-  type?:
-    | "checkbox"
-    | "button"
-    | "color"
-    | "date"
-    | "datetime-local"
-    | "email"
-    | "file"
-    | "hidden"
-    | "image"
-    | "month"
-    | "number"
-    | "password"
-    | "radio"
-    | "range"
-    | "reset"
-    | "search"
-    | "submit"
-    | "tel"
-    | "text"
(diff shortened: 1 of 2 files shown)


src/form/input/Input.tsx
```diff
@@ -8,6 +8,7 @@ export type InputProps = Omit<
   "disabled" | "name"
 > & {
   name: string;
+  type?: "text" | "color";
```
edizcelik, line 11:
```suggestion
  type?:
    | "checkbox"
    | "button"
    | "color"
    | "date"
    | "datetime-local"
    | "email"
    | "file"
    | "hidden"
    | "image"
    | "month"
    | "number"
    | "password"
    | "radio"
    | "range"
    | "reset"
    | "search"
    | "submit"
    | "tel"
    | "text"
    | "time"
    | "url"
    | "week";
```

Let's provide all available options

src/form/input/Input.tsx
```diff
@@ -8,6 +8,7 @@ export type InputProps = Omit<
   "disabled" | "name"
```
edizcelik, line 8: Let's omit `className` too so that users of the component are forced to use `customClassName` instead
```suggestion
  "disabled" | "name" | "className"
```

edizcelik (changes requested):
Let's clean the implementation of Input component as well. Some of the variables we destruct and assign to a const within the body of the component is not necessary. For example, `step`, `min`, `max` are not used anywhere else within the component but just provided to the `input` as prop. There is no point in destructuring those. We should only destructure values that we are actually using or assigning them a different default value:

```
  const {
    testid,
    type = "text",
    isDisabled,
    hasError,
    customClassName,
    leftIcon,
    rightIcon,
    role,
    autoComplete = "off",
    autoCorrect = "off",
    inputContainerRef,
    ...rest
  } = props;
  
  
  ...
  
  
        <input
        className={inputClassName}
        type={type}
        placeholder={placeholder}
        autoComplete={autoComplete}
        autoCorrect={autoCorrect}
        disabled={isDisabled}
        {...rest}
      />
```

edizcelik (approved):

jamcry (approved):

bilgeyucel (approved):

Anlerkan (approved):

src/form/input/Input.tsx
```diff
@@ -5,9 +5,32 @@ import classNames from "classnames";
 
 export type InputProps = Omit<
   React.InputHTMLAttributes<HTMLInputElement>,
-  "disabled" | "name"
+  "disabled" | "name" | "className"
 > & {
   name: string;
+  type?:
+    | "checkbox"
```
mucahit, line 12: What do you think about creating a separate type for the input types?

mucahit (approved):

PR #64, opened 2021-04-07
feat(textarea): Add Textarea component

yasincaliskan:
### Description
- Install `react-textarea-autosize` package `8.3.2` exact version.
- Textarea implementation for fixed and auto size.

### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/33

image


src/form/textarea/Textarea.tsx
```diff
@@ -0,0 +1,153 @@
+import "./_textarea.scss";
+
+import React, {useState} from "react";
+import classNames from "classnames";
+import TextareaAutosize, {TextareaAutosizeProps} from "react-textarea-autosize";
+
+import {KEYBOARD_EVENT_KEY} from "./util/textareaConstants";
```
Anlerkan, line 7: I think we can use `KEYBOARD_EVENT_KEY` from `"../../core/utils/keyboardEventConstants"` here.

yasincaliskan: You're right, I've missed this. Thank you!

src/form/textarea/_textarea.scss
```diff
@@ -0,0 +1,49 @@
...
+    padding: 0;
+
+    background-color: transparent;
+    border: none;
+    border-bottom: 1px solid var(--default-border-color);
+    border-radius: 0;
+
+    transition: border-bottom-color var(--default-animation-duration)
+      var(--default-animation-function);
+  }
```
edizcelik, line 48: I don't think we need this on the toolkit.

package.json
```diff
@@ -56,6 +56,7 @@
     "react-dom": "16.13.1",
     "react-svg-loader": "3.0.3",
     "react-test-renderer": "16.13.1",
+    "react-textarea-autosize": "8.3.2",
```
edizcelik, line 59: I think this should be a peer dependency.

src/form/textarea/Textarea.tsx
```diff
@@ -0,0 +1,153 @@
...
+    autoFocus = false,
+    ...otherProps
+  } = props;
+  const [isShiftPressed, setShiftPressedState] = useState(false);
+  const containerClassName = classNames(
+    "textarea-container",
+    customClassNames?.container
+  );
+  const textareaClassName = classNames("textarea", customClassNames?.textarea, {
+    "textarea--has-error": hasError
```
edizcelik, line 61: Let's add this class to the container. Let's add a `is-disabled` modifier class to the container if the textarea is disabled as well

stories/0-Form.stories.tsx
```diff
@@ -57,6 +59,49 @@ storiesOf("Form", module)
...
+      <Textarea
+        id={"textarea-auto-size"}
+        name={"Disabled auto size Textarea"}
+        placeholder={"Write a paragraph"}
+        autoSizeProps={{}}
+        isDisabled={true}
+        onChange={(e) => {
+          console.log(e.currentTarget.value);
+        }}
+      />
```
edizcelik, line 102: Let's add an example with `hasError`

src/form/textarea/Textarea.tsx
```diff
@@ -0,0 +1,153 @@
...
+  style?: React.CSSProperties;
+  autoSizeProps?: TextareaAutosizeProps;
+  onJustEnterPressed?: () => void;
+  onShiftEnter?: () => void;
+  isRequired?: boolean;
+  hasError?: boolean;
+};
+
+function Textarea(props: TextareaProps) {
+  const {
```
edizcelik, line 31: Let's try to clean this here and only destructure the variables that are necessary. If all the variables are required to destructure, you can ignore this however as far as I can see we can remove some of them and spread `otherProps` when passing props to the elements below

edizcelik (changes requested):

src/form/textarea/Textarea.tsx
```diff
@@ -0,0 +1,137 @@
...
+          autoFocus={autoFocus}
+          {...otherProps}
+        />
+      )}
+    </div>
+  );
+
+  function handleKeyDown(event: React.KeyboardEvent<HTMLTextAreaElement>) {
+    const {key: pressedKey} = event;
+    const {onKeyDown} = props;
```
jamcry, line 102: I think we can destructure `onKeyDown` above with other props (just like you did with `onKeyUp`) and remove this line. It may be cleaner because currently `props` is destructured twice
```suggestion
```
```ts
...
   onKeyDown,
    ...otherProps
  } = props;
```

src/form/textarea/Textarea.tsx
```diff
@@ -0,0 +1,137 @@
...
+    autoFocus = false,
+    ...otherProps
+  } = props;
+  const [isShiftPressed, setShiftPressedState] = useState(false);
+  const containerClassName = classNames(
+    "textarea-container",
+    customClassNames?.container,
+    {"textarea--has-error": hasError,
+     "textarea--is-disabled" : isDisabled
+    }
```
jamcry, line 55: Is this formatting valid? 🤔  I think it should be like this (according to our linter rules):
```suggestion
    {
     "textarea--has-error": hasError,
     "textarea--is-disabled" : isDisabled
    }
```

src/form/textarea/Textarea.tsx
```diff
@@ -0,0 +1,137 @@
...
+  testid?: string;
+  isDisabled?: boolean;
+  customClassNames?: {
+    container?: string;
+    textarea?: string;
+  };
+  style?: React.CSSProperties;
+  autoSizeProps?: TextareaAutosizeProps;
+  onJustEnterPressed?: () => void;
+  onShiftEnter?: () => void;
```
jamcry, line 25: I think both are OK, but I usually prefer `VoidFunction` (a pre-defined type) instead of `() => void`. They represent the same thing, but it is easier to type `VoidFunction` 😄

jamcry (commented):
Looks good, thanks! 👏  I have just a few suggestions

src/form/textarea/Textarea.tsx
yasincaliskan: Yes, looks good. So, should we refactor all of the `() => void` usage to `VoidFunction`? It's just for consistency.

src/form/textarea/Textarea.tsx
yasincaliskan: You're right. I guess we should update our `eslint & stylelint` packages. Can we @edizcelik ?

src/form/textarea/Textarea.tsx
jamcry: I don't think we should, if `() => void` is more common in this codebase, just wanted to let you know about this option. It can stay like this 👍

yasincaliskan: Alright, I'm reverting the related change. Thank you for your comments! 🎉

bilgeyucel (approved):

jamcry (approved):

Anlerkan (approved):

mucahit (approved):

src/form/textarea/Textarea.tsx
edizcelik: No this should be handled by Prettier

edizcelik (approved):

PR #66, opened 2021-04-09
feat(toast): Add `Toast` component

yasincaliskan:
### Description
- React versions have upgraded to `17.0.2` for hooks to work properly.
- ToastDecorator has added with `addDecorator`. Thus, ToastProvider can wrap global Story.
- Toast sample has created out of the `storiesOf` function for can use `useToaster` hook.

### Notes
- We can upgrade Storybook version to the latest version (6.7.2) after these PR's merged. Because there are some breaking changes on the config files.

### Fixes:
- https://github.com/Hipo/react-ui-toolkit/issues/61


edizcelik (changes requested):
We should export Toast related stuff from index so that it is available to use on other projects

src/toast/util/toastHooks.ts
```diff
@@ -0,0 +1,13 @@
...
+import {ToastContext} from "../ToastProvider";
+
+function useToast() {
+  return useContext(ToastContext);
+}
+
+function useToaster() {
+  return useToast().dispatchToastAction;
+}
+
```
jamcry, line 12: I think adding a hook for directly showing toast would be great, because we mostly use "DISPLAY" action with autoClose feature, and for each display action we have to pass `type` and `payload`, this creates too much duplicated code IMO

```suggestion
function useDisplayToast() {
  const dispatchToast =  useToaster();
  
  function displayToast(payload:ToastData) {
    dispatchToast({
      type: "DISPLAY",
      payload
    });
  }
  
  return displayToast;
}
```

This would reduce:
```ts
dispatchToast({
  type: "DISPLAY",
  payload: {
    mode: "success",
    component: "Success Message"
  }
})
```
   to this:
```ts
displayToast({
   mode: "success",
   component: "Success Message"
})
```

This seems a little change but overall it can help reduce duplicated code.
Maybe we can discuss this in our weekly meeting

src/toast/util/toastTypes.ts
```diff
@@ -0,0 +1,16 @@
+interface ToastData {
+  mode: "danger" | "success";
+  component: React.ReactNode;
```
jamcry, line 3: I think `content` can be a more descriptive name for this, "content of the toast" makes more sense to me. What do you think?

src/toast/Toast.tsx
```diff
@@ -0,0 +1,43 @@
...
+    };
+  }, [isDisplayed, autoClose, timeout, dispatchToastAction]);
+
+  const toast = (
+    <div className={classNames("toast", `toast--${mode}`, customClassName)}>
+      {component}
+    </div>
+  );
+
+  return ReactDOM.createPortal(toast, document.querySelector("#toast-root")!);
```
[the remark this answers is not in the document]
yasincaliskan: `react-modal` is using `body` for display modal. It makes a new div selector and using this as [default](https://github.com/reactjs/react-modal/blob/d7083c5303775450ff27063857e61eb5b653e418/src/components/Modal.js). We could think it is `<div id="#root" />`. As far as I understand the modal selectors exist on runtime and removing the child selector after.

src/toast/util/toastTypes.ts
yasincaliskan: Sounds good. We can change it as `content`.

src/toast/_toast.scss
```diff
@@ -0,0 +1,20 @@
+.toast {
+  position: fixed;
+  bottom: 0;
+  z-index: 2;
```
bilgeyucel, line 4: I think we can make the `z-index` a very large number, such as `1000`, so that it will be above all content.

yasincaliskan: We decided we can handle this on project scope. I think the initial `z-index` value can stay 2.

src/toast/Toast.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  if (!toastRootNode) {
+    toastRootNode = document.createElement("div");
+    toastRootNode.setAttribute("id", "toast-root");
+  }
+
+  document.body.appendChild(toastRootNode);
+
+  useEffect(() => {
+    let timeoutId: number;
```
jamcry, line 27: I think `NodeJS.Timeout` is the correct type for this variable

(see return type in the ss)
Screen Shot 2021-04-14 at 12 48 01 PM

jamcry (approved):
Other than my one comment, looks good to go 👍

src/toast/Toast.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+function Toast() {
+  const {
+    toastState: {
+      isDisplayed,
+      data: {autoClose, timeout, content, mode, customClassName}
+    },
+    dispatchToastAction
+  } = useToast();
+  let toastRootNode = document.querySelector("#toast-root");
```
mucahit, line 17: Let's add a `toastRootId` prop
```suggestion
  let toastRootNode = document.querySelector(toastRootId || "#toast-root");
```

src/toast/Toast.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+  if (!toastRootNode) {
+    toastRootNode = document.createElement("div");
+    toastRootNode.setAttribute("id", "toast-root");
+  }
+
+  document.body.appendChild(toastRootNode);
```
mucahit, line 24: Shouldn't it be inside of the `if(!toastRootNode)` check?

src/toast/_toast.scss
```diff
@@ -0,0 +1,20 @@
...
+
+  width: 100vw;
+
+  padding: 17px 0;
+
+  text-align: center;
+
+  &--danger {
+    color: red;
```
mucahit, line 13: Let's make it CSS variable

`--toast-danger-text-color`
`--toast-danger-bg`

`--toast-warning-text-color`
`--toast-warning-bg`

`--toast-success-text-color`
`--toast-success-bg`

src/toast/util/toastTypes.ts
```diff
@@ -0,0 +1,16 @@
+interface ToastData {
+  mode: "danger" | "success";
```
mucahit, line 2: Let's add `warning` type

src/toast/util/toastTypes.ts
```diff
@@ -0,0 +1,16 @@
+interface ToastData {
```
mucahit, line 1: Maybe?
```suggestion
interface ToastItem {
```

mucahit (changes requested):

src/toast/Toast.tsx
yasincaliskan: If doesn't exist a selector `#toast-root`, we are creating one inside of the `if(!toastRootNode)`, if already exists we just append it to the body. I think we should add for each two situations.

src/index.ts
```diff
@@ -63,5 +66,8 @@ export {
   FileUploadButtonProps,
   SpinnerProps,
   TabItem,
-  TabProps
+  TabProps,
+  ToastProps,
+  useToaster,
+  useDisplayToast
```
edizcelik, line 72: These aren't types. Let's move these two up to come after Toast. We should export `useToast` hook as well. We also have to export ToastContext and ToastContextProvider.

src/toast/util/toastHooks.ts
```diff
@@ -0,0 +1,27 @@
...
+function useDisplayToast() {
```
edizcelik, line 14: Let's add JSDoc to describe what each of these hooks does

src/toast/Toast.tsx
mucahit: I didn't get it. If it already exists why we need to append it?

yasincaliskan: You are absolutely right. I don't remember what I thought 😅

src/toast/Toast.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    if (autoClose) {
+      timeoutId = setTimeout(() => {
+        dispatchToastAction({
+          type: "HIDE"
+        });
+
+        if (toastRootNode) {
+          document.body.removeChild(toastRootNode);
+        }
```
bilgeyucel, line 40: Don't we remove the `toastRootNode` the user provided (not generated with line 26) with this code? 🤔  I guess we wouldn't want that

yasincaliskan: Which is `toastRootNode` that the user provided? In this case, we just create a node and then remove this after display the message.

edizcelik: Bilge is right. I think we should keep the node if it was provided by the user of the component. BTW, how do I provide `toastRootId` to this component when integrating it to a project ? Also, DOM manipulations within this component should be done within a layoutEffect, and the removal of the node can be handled within that effect's cleanup.

edizcelik:
@yasincaliskan you can check Toast related files within this PR: https://github.com/Hipo/algoswap-web-frontend/pull/7

src/toast/ToastProvider.tsx
```diff
@@ -0,0 +1,34 @@
...
+import toastReducer from "./util/toastReducer";
+import {ToastAction, ToastContextState} from "./util/toastTypes";
+
+const ToastContext = createContext<[ToastContextState, React.Dispatch<ToastAction>]>([
+  initialToastState,
+  () => undefined
+]);
+
+interface ToastContextProviderProps {
+  children: React.ReactNode;
```
edizcelik, line 14: Currently we don't have a way to provide `customRootId` to ToastStack component. I think we can add that as a prop to the `ToastContextProvider` and users can provide the customRootId from there.

src/toast/stack/_toast-stack.scss
```diff
@@ -0,0 +1,10 @@
+.toast-stack {
+  position: fixed;
+  bottom: 0;
+  left: calc(50vw - var(--toast-width) / 2);
+
+  display: flex;
+  flex-direction: column-reverse;
+  align-items: center;
+  justify-content: center;
```
edizcelik, line 9: Let's just display them on the bottom right by default
```suggestion
  position: fixed;
  bottom: 0;
  right: 0;
  z-index: 2;

  max-height: 100vh;

  overflow: auto;

  padding: 24px;
```

src/toast/util/toastHooks.ts
```diff
@@ -0,0 +1,77 @@
...
+
+import {generateRandomString} from "../../core/utils/string/stringUtils";
+import {ToastContext} from "../ToastProvider";
+import {ToastData} from "./toastTypes";
+
+/**
+ * @returns {Object} Current value of ToastContext
+ */
+function useToastContext() {
+  return useContext(ToastContext);
```
edizcelik, line 11: Let's throw an error if the context is being used outside of a Provider component tree

src/toast/util/toastTypes.ts
```diff
@@ -0,0 +1,25 @@
+export interface ToastData {
+  mode: "danger" | "warning" | "success" | "info";
```
edizcelik, line 2: Let's not have any `mode` as we don't provide any default UI within Toast component.

src/toast/util/toastTypes.ts
```diff
@@ -0,0 +1,25 @@
...
+  content: React.ReactNode;
```
edizcelik, line 3: Let's call it a render and it should be a function
```suggestion
  render: () => React.ReactNode;
```

src/ui/reference/_measurement.scss
```diff
@@ -1,4 +1,6 @@
 :root {
   // Defaults
   --small-radius-size: 2px;
+
+  --toast-width: 300px;
```
edizcelik, line 5: Let's define this within `_toast-stack.scss` or `_toast.scss`

src/toast/_toast.scss
```diff
@@ -0,0 +1,58 @@
+.toast {
```
edizcelik, line 1: We can remove these styles altogether I think

stories/8-Toast.stories.tsx
```diff
@@ -0,0 +1,96 @@
...
+              timeout: 10000
+            })
+          }>
+          {"Update Custom Toast"}
+        </Button>
+
+        <Button type={"button"} onClick={hideAll}>
+          {"Hide All"}
+        </Button>
+      </div>
```
edizcelik, line 91: Let's add an example where the toast doesn't `autoClose` but it is controlled by a button within the provided "render" function to the toast.display function. I think the way we will have this functionality is to have a `ToastCloseButton` component which as access to that particular Toast's id so that it can close it. This will be another compound components pattern. Each Toast component should provide a context to their children and this ToastCloseButton can then access this context to use Toast's id. We will export ToastCloseButton from ui-toolkit as well so that it can be used within projects.

src/toast/_toast.scss
yasincaliskan: I moved it under the `stories` directory to display colorful. Can it stay like thatthis? 
Also, CSS variable declaration can help create Toast component in the project.

stories/8-Toast.stories.tsx
yasincaliskan: I've added an example and `ToastStory` component to use compound component pattern. Should we create like that?

src/toast/Toast.tsx
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    return () => {
+      clearTimeout(timeoutId);
+    };
+  }, [autoClose, timeout, hide, toastId]);
+
+  return (
+    <ListItem testid={testid} customClassName={customClassName}>
+      {render()}
```
edizcelik, line 39: We need to wrap with a context here so that any component defined under `render` can reach to this context to use, for example, this toast's id.

stories/8-Toast.stories.tsx
edizcelik: No. See my comment below. We need to provide this functionality to the users of this toolkit

src/toast/Toast.tsx
yasincaliskan: Thank you for your advice 🙏🏻

src/toast/Toast.tsx
```diff
@@ -0,0 +1,50 @@
...
+      }, timeout);
+    }
+
+    return () => {
+      clearTimeout(timeoutId);
+    };
+  }, [autoClose, timeout, hide, toastId]);
+
+  return (
+    <ToastCloseButtonContext.Provider value={{toastId}}>
```
edizcelik, line 40:
```suggestion
    <ToastItemContext.Provider value={{toastId}}>
```

I think this is a better name

src/toast/close-button/ToastCloseButtonProvider.tsx
```diff
@@ -0,0 +1,23 @@
+import {createContext, useContext} from "react";
```
edizcelik, line 1: Let's rename this file as `ToastItemContext.tsx`

src/toast/close-button/ToastCloseButtonProvider.tsx
```diff
@@ -0,0 +1,23 @@
...
+  toastId: ""
+});
+
+ToastCloseButtonContext.displayName = "ToastCloseButtonContext";
+
+function useToastCloseButton() {
+  const toastStoryContext = useContext(ToastCloseButtonContext);
+
+  if (!toastStoryContext) {
+    throw new Error("No context found for ToastCloseButtonContext");
```
edizcelik, line 17:
```suggestion
    throw new Error("Trying to consume ToastItemContext outside of its provider");
```

src/toast/close-button/_toast-close-button.scss
```diff
@@ -0,0 +1,5 @@
+.toast-close-button {
+  --toast-close-button-z-index: 3;
+  
+  z-index: var(--toast-close-button-z-index);
+}
```
edizcelik, line 5: No need for the css variable. I think we should rethink the use of extensive css variables. Many of them within toolkit are actually redundant because we can easily override the styles within projects now with CSS.

src/toast/stack/_toast-stack.scss
```diff
@@ -0,0 +1,15 @@
+.toast-stack {
+  --toast-width: 300px;
+  --toast-z-index: 2;
```
edizcelik, line 3: Same here. Let's remove the variables

src/toast/util/toastHooks.ts
```diff
@@ -0,0 +1,83 @@
...
+import {ToastContext} from "../ToastProvider";
+import {ToastData} from "./toastTypes";
+
+/**
+ * @returns {Object} Current value of ToastContext
+ */
+function useToastContext() {
+  const context = React.useContext(ToastContext);
+
+  if (context === undefined) {
```
edizcelik, line 13:
```suggestion
  if (!context) {
```

src/toast/util/toastHooks.ts
```diff
@@ -0,0 +1,83 @@
...
+    throw new Error("No context found for ToastContext");
```
edizcelik, line 14:
```suggestion
    throw new Error("Trying to consume ToastContext outside of its provider.");
```

stories/8-Toast.stories.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const {display, update, hideAll} = useToaster();
+
+  return (
+    <StoryFragment>
+      <div className={"toast-button-group"}>
+        <Button
+          type={"button"}
+          onClick={() =>
+            display({
+              render: () => {
```
edizcelik, line 21: Same for others below too
```suggestion
              render() {
```

src/toast/util/toastHooks.ts
yasincaliskan: Can we update Toggle Error message also like this?

src/toast/close-button/_toast-close-button.scss
yasincaliskan: Maybe we can think we need which variables and then we can remove.

src/toast/util/toastHooks.ts
edizcelik: Yep

stories/utils/constants/toast/_toast.scss
```diff
@@ -0,0 +1,70 @@
...
+  --toast-success-text-color: white;
+  --toast-success-bg: yellowgreen;
+
+  --toast-info-text-color: white;
+  --toast-info-bg: darkcyan;
+
+  --toast-height: 20px;
+  --toast-margin-bottom: 10px;
+  --toast-margin-top: 10px;
+  --toast-z-index: 2;
```
edizcelik, line 17: Let's not use css variables here

src/toast/close-button/ToastItemContext.tsx
```diff
@@ -0,0 +1,23 @@
...
+  toastId: string;
+};
+
+const ToastItemContext = createContext<ToastItemContextValue>({
+  toastId: ""
+});
+
+ToastItemContext.displayName = "ToastItemContext";
+
+function useToastCloseButton() {
```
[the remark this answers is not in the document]
yasincaliskan: All hooks in together would be better 👍🏻

src/toast/util/toastHooks.ts
```diff
@@ -0,0 +1,94 @@
...
+     */
+    hideAll: useCallback(() => {
+      dispatch({
+        type: "HIDE_ALL"
+      });
+    }, [dispatch])
+  };
+}
+
+function useToastCloseButton() {
```
edizcelik, line 84:
```suggestion
function useToastItemContext() {
```

src/toast/close-button/ToastItemContext.tsx
```diff
@@ -0,0 +1,13 @@
+import {createContext} from "react";
```
edizcelik, line 1: It doesn't make sense to have this file under `close-button` directory. Let's move it to `src/toast/` or `src/toast/util/`

edizcelik (changes requested):

src/toast/close-button/ToastItemContext.tsx
yasincaliskan: You're right. `src/toast` is better.

bilgeyucel (approved):

jamcry (approved):

src/toast/util/toastUtils.ts
```diff
@@ -0,0 +1,11 @@
+import {ToastData} from "./toastTypes";
+
+function isDifferentToast(toastId: string) {
+  return (toast2: ToastData) => toastId !== toast2.id;
+}
+
+function isSameToast(toastId: string) {
+  return (toast2: ToastData) => toastId === toast2.id;
+}
+
```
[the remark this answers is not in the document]
edizcelik: We can keep one of them

edizcelik (approved):

PR #70, opened 2021-04-12
feat(progress-bar): Add ProgressBar Component

yasincaliskan:
### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/68

### DIFF
--- a/src/progress-bar/ProgressBar.tsx
+++ b/src/progress-bar/ProgressBar.tsx
@@ -10,11 +10,7 @@ export interface ProgressBarProps {
     backgroundColor?: string;
     completedColor?: string;
   };
-  children: {
-    content?: React.ReactNode;
-    position?: "top" | "bottom" | "inside";
-  };
-  view?: "line" | "bar";
+  children?: React.ReactNode;
   customClassName?: string;
 }
 
@@ -22,25 +18,18 @@ function ProgressBar(props: ProgressBarProps) {
   const {
     percentage,
     style: {trackColor = "blue", backgroundColor = "gray", completedColor = "green"},
-    children: {content, position = "top"},
-    view = "bar",
+    children,
     customClassName
   } = props;
   const progressBarContainerClassName = classNames(
     "progress-bar-container",
     customClassName
   );
-  const progressBarClassName = classNames("progress-bar", {
-    "progress-bar--is-line-view": view === "line",
-    "progress-bar--is-bar-view": view === "bar"
-  });
 
   return (
     <div className={progressBarContainerClassName}>
-      {content && position !== "inside" && renderContent()}
-
       <div
-        className={progressBarClassName}
+        className={"progress-bar"}
         style={{
           backgroundColor: backgroundColor
         }}>
@@ -52,22 +41,10 @@ function ProgressBar(props: ProgressBarProps) {
           }}
         />
 
-        {content && position === "inside" && view === "bar" && renderContent()}
+        {children && <div className={"progress-bar__content"}>{children}</div>}
       </div>
     </div>
   );
-
-  function renderContent() {
-    return (
-      <div
-        className={classNames(
-          "progress-bar__content",
-          `progress-bar__content--is-position-${position}`
-        )}>
-        {content}
-      </div>
-    );
-  }
 }
 
 export default ProgressBar;
--- a/src/progress-bar/ProgressBar.tsx
+++ b/src/progress-bar/ProgressBar.tsx
@@ -21,28 +21,23 @@ function ProgressBar(props: ProgressBarProps) {
     children,
     customClassName
   } = props;
-  const progressBarContainerClassName = classNames(
-    "progress-bar-container",
-    customClassName
-  );
+  const progressBarClassName = classNames("progress-bar", customClassName);
 
   return (
-    <div className={progressBarContainerClassName}>
+    <div
+      className={progressBarClassName}
+      style={{
+        backgroundColor: backgroundColor
+      }}>
       <div
-        className={"progress-bar"}
+        className={"progress-bar__track"}
         style={{
-          backgroundColor: backgroundColor
-        }}>
-        <div
-          className={"progress-bar__track"}
-          style={{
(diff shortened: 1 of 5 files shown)


bilgeyucel (approved):

jamcry (approved):

Anlerkan (approved):

src/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,72 @@
...
+interface ProgressBarProps {
+  percentage: number;
+  style: {
+    trackColor?: string;
+    backgroundColor?: string;
+    completedColor?: string;
+  };
+  children: {
+    content?: React.ReactNode;
+    position?: "top" | "bottom" | "in";
```
mucahit, line 15:
```suggestion
    position?: "top" | "bottom" | "inside";
```

src/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,72 @@
...
+  );
+  const progressBarClassName = classNames("progress-bar", {
+    "progress-bar--is-line-view": view === "line",
+    "progress-bar--is-bar-view": view === "bar"
+  });
+
+  return (
+    <div className={progressBarContainerClassName}>
+      {content && position !== "in" && (
+        <div
```
mucahit, line 41: Let's make it a method `renderContent` to avoid duplicate JSX usage

mucahit (changes requested):

src/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,73 @@
...
+  percentage: number;
+  style: {
+    trackColor?: string;
+    backgroundColor?: string;
+    completedColor?: string;
+  };
+  children: {
+    content?: React.ReactNode;
+    position?: "top" | "bottom" | "inside";
+  };
```
edizcelik, line 16: I have never seen children to be an object instead of a React node and it looks weird. I don't think this is a good practice. It doesn't make sense to handle the positioning here. If children is provided it should be "inside" the track, "top" and "bottom" should be handled on the call-site. These are the use cases:

```
<ProgressBar>{"50% - appears within the ProgressBar"}</ProgressBar>

...

<div>
   <p>{"50% - some content above ProgressBar"}</p>
   <ProgressBar />
<div>

...

<div>
   <ProgressBar />
   <p>{"50% - some content below ProgressBar"}</p>
<div>
``` 

You can add examples of this to Storybook

src/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,73 @@
...
+  view?: "line" | "bar";
```
edizcelik, line 17: Let's remove this prop too. They can control the height with CSS.

src/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,73 @@
...
+          className={"progress-bar__track"}
+          style={{
+            width: `${percentage}%`,
+            backgroundColor: percentage === 100 ? completedColor : trackColor
+          }}
+        />
+
+        {content && position === "inside" && view === "bar" && renderContent()}
+      </div>
+    </div>
```
edizcelik, line 57: As we shouldn't handle the positioning of the content here, the DOM structure should look like this:

```
<div className={"progress-bar"}>
   <div className={"progress-bar__track"} />
   
   {children && <div className={"progress-bar__content"}>{children}</div>}
</div>
```

edizcelik (changes requested):

stories/9-ProgressBar.stories.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+const progressBarStory = (
+  <style>{`
+  .progress-bar-story {
+    height: 51.2px;
+  }
+`}</style>
+);
+
+storiesOf("Progress Bar", module)
```
edizcelik, line 20: Stories added here are misleading. Storybook example should just demonstrate the use of different available props to the component. So we can have the following examples:

```
ProgressBar with different percentages
ProgressBar with height overridden by CSS
ProgressBar with children
```

There aren't any children on "top", or "bottom". Those aren't children for the ProgressBar component and is irrelevant to the component itself.

stories/9-ProgressBar.stories.tsx
```diff
@@ -0,0 +1,206 @@
...
+      <br />
+
+      <ProgressBar
+        percentage={85}
+        style={{
+          backgroundColor: colors.background,
+          trackColor: colors.track,
+          completedColor: colors.completed
+        }}
+        children={<p>{"85 %"}</p>}
```
edizcelik, line 189: Let's follow the usual convention for passing children to React elements:

```jsx
<ProgressBar>{<p>{"85 %"}</p>}</ProgressBar>
```

edizcelik (changes requested):

stories/9-ProgressBar.stories.tsx
```diff
@@ -0,0 +1,170 @@
...
+        percentage={50}
+        style={{
+          backgroundColor: colors.background,
+          trackColor: colors.track,
+          completedColor: colors.completed
+        }}
+        customClassName={"progress-bar-story"}
+      />
+
+      <p>{"50 %"}</p>
```
edizcelik, line 130: We should remove these two examples because they don't pass any `children` to ProgressBar

edizcelik (changes requested):

src/progress-bar/ProgressBar.tsx
```diff
@@ -0,0 +1,50 @@
...
+    children,
+    customClassName
+  } = props;
+  const progressBarContainerClassName = classNames(
+    "progress-bar-container",
+    customClassName
+  );
+
+  return (
+    <div className={progressBarContainerClassName}>
```
edizcelik, line 30: Let's remove this redundant wrapper div and any related styles attached to it on the SCSS file

edizcelik (approved):

mucahit (approved):

PR #71, opened 2021-04-12
feat(number-input): Add NumberInput Component

yasincaliskan:
### Description
- `NumberInput` component has added.
- Add Number utilities have added.
- Core utilities have grouped.
- `shouldFormatToLocaleString` has removed.

### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/69


jamcry (approved):

src/dropdown/Dropdown.tsx
```diff
@@ -6,7 +6,7 @@ import classNames from "classnames";
 
 import Button from "../button/Button";
 import DropdownList from "./list/DropdownList";
-import {KEYBOARD_EVENT_KEY} from "../core/utils/keyboardEventConstants";
+import {KEYBOARD_EVENT_KEY} from "../core/utils/keyboard-event/keyboardEventConstants";
```
mucahit, line 9: I guess `../core/utils/keyboard/keyboardEventConstants` should be enough

src/form/input/number/NumberInput.tsx
```diff
@@ -0,0 +1,157 @@
...
+  hasError,
+  placeholder,
+  customClassName,
+  leftIcon,
+  onFocus,
+  onKeyDown,
+  prefixIconToValue,
+  isDisabled
+}: NumberInputProps) {
+  const className = classNames("number-input", customClassName);
```
mucahit, line 52:
```suggestion
  const className = classNames("input__number", customClassName);
```

mucahit (changes requested):

src/form/input/number/NumberInput.tsx
edizcelik: This is a block on its own and its not under "input" classname block so it should actually be `number-input` or we could think of it as a modifier instead of a block: `input--number-type`

src/form/input/number/NumberInput.tsx
```diff
@@ -0,0 +1,157 @@
...
+  isDisabled
+}: NumberInputProps) {
+  const className = classNames("input__number", customClassName);
+  const containerRef = useRef() as RefObject<HTMLDivElement>;
+
+  let finalValue = value;
+
+  if (finalValue && prefixIconToValue && !finalValue.includes(prefixIconToValue)) {
+    finalValue = `${prefixIconToValue} ${finalValue}`;
+  }
```
edizcelik, line 59: This is not a good UX but was hack used in a project. I think we should remove this from the toolkit and have a proper UX solution for the number inputs that require displaying a symbol/unit. This require assistance from the design team so let's just remove the prop for now.

src/form/input/number/NumberInput.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+    let formattedNewValue = newValue
+      .replace(NOT_INTEGER_FIRST_CHARACTER_OF_STRING_REGEX, "")
+      .replace(NOT_NUMBER_NOR_DECIMAL_POINT_REGEX, "")
+      .replace(PRECISION_REGEX, "$1");
+
+    if (maxFractionDigits > 0) {
+      const decimalNumberParts = newValue.split(DECIMAL_NUMBER_SEPARATOR);
+      const decimalPart = decimalNumberParts[1];
+      const integerPart = decimalNumberParts[0].replace(THOUSANDTHS_SEPARATOR, "");
```
edizcelik, line 135: These can stay for now but we will need to rethink these while we are working on the localization because `DECIMAL_NUMBER_SEPARATOR` and `THOUSANDTHS_SEPARATOR` changes for different locales

edizcelik (changes requested):
Didn't we agree that this NumberInput will be an abstraction of Input component when the type is provided as "number" ? Meaning Input component will use NumberInput behind the scenes if `type === "number"`

src/core/utils/number/numberUtils.ts
```diff
@@ -0,0 +1,37 @@
...
+  DECIMAL_NUMBER_SEPARATOR
+} from "./numberConstants";
+
+/**
+ * Coerces a number into a string.
+ * @param {number} value A number to convert to string
+ * @param {string} locale Default locale used is "en"
+ * @return {string} The value after coercing the given value to a string.
+ */
+function numberToString(value: unknown, locale = "en") {
```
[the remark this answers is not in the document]
yasincaliskan: Alright. I have an idea for localization. I'll open a new PR after these PR's merged.

src/form/input/Input.tsx
```diff
@@ -77,10 +89,11 @@ function Input(props: InputProps) {
 
       <input
         className={inputClassName}
-        type={type}
+        type={isNumberInput ? "text" : type}
```
edizcelik, line 92: As we discussed, NumberInput component should still be there but `Input` component should just render `NumberInput` if `type === "number"` and pass all of the props to the `NumberInput`

edizcelik (changes requested):

src/form/input/Input.tsx
yasincaliskan: Oh, I get it. `NumberInput` component should exist but using only by `Input` component. 👍🏻

edizcelik: Just to give update for others, as NumberInput also uses Input, this wouldn't work and leads to an infinite loop. Therefore, after talking with Yasin, we decided that it would be better to remove NumberInput altogether and just keep Input component. We will do specific calculations within Input when `type` props is equal to "number".

bilgeyucel (approved):
🚀

stories/0-Form.stories.tsx
```diff
@@ -135,6 +135,65 @@ storiesOf("Form", module)
...
+          </Fragment>
+        )}
+      </StateProvider>
+
+      <br />
+
+      <StateProvider initialState={null}>
+        {(state, setState) => (
+          <Fragment>
+            <FormField label={"Price (BTC)"}>
```
edizcelik, line 181:
```suggestion
            <FormField label={"Price (BTC) – maxFractionDigits={8}"}>
```

stories/0-Form.stories.tsx
```diff
@@ -135,6 +135,65 @@ storiesOf("Form", module)
       />
     </FormField>
   ))
+
+  .add("Number Input", () => (
+    <Fragment>
+      <StateProvider initialState={null}>
+        {(state, setState) => (
+          <Fragment>
+            <FormField label={"Price"}>
```
edizcelik, line 144:
```suggestion
            <FormField label={"Price – maxFractionDigits={2}"}>
```

stories/0-Form.stories.tsx
```diff
@@ -135,6 +135,65 @@ storiesOf("Form", module)
       />
     </FormField>
   ))
+
+  .add("Number Input", () => (
```
edizcelik, line 139: Let's add an example where `maxFractionDigits` is 0 so that only integers are allows

edizcelik (approved):

Anlerkan (approved):

mucahit (approved):

PR #72, opened 2021-04-12
feat(countdown): Add Countdown component

yasincaliskan:
### Description
- Countdown component and time utilities have added

### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/67


src/countdown/Countdown.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+function Countdown({
+  testid,
+  startDate,
+  onEnd,
+  countDownIntervalInSeconds = 1,
+  customClassName
+}: CountdownProps) {
+  const countDownData = useCountDownTimer(startDate, countDownIntervalInSeconds);
+  let items: {id: "days" | "hours" | "minutes" | "seconds"; count: string | number}[];
```
bilgeyucel, line 26: You can define a new type `CountDownItem` in this file, remove else from the conditions and give it to `items` as the first value.

```suggestion
  let items: CountDownItem[] = [
      {id: "minutes", count: countDownData.minutes},
      {id: "seconds", count: countDownData.seconds}
    ];
```

```
interface CountDownItem {
    id: "days" | "hours" | "minutes" | "seconds";
    count: string | number;
}
```

src/countdown/Countdown.tsx
```diff
@@ -0,0 +1,72 @@
...
+    items = [
+      {id: "days", count: countDownData.days},
+      {id: "hours", count: countDownData.hours},
+      {id: "minutes", count: countDownData.minutes}
+    ];
+  } else if (countDownData.hours >= 1) {
+    items = [
+      {id: "hours", count: countDownData.hours},
+      {id: "minutes", count: countDownData.minutes}
+    ];
```
bilgeyucel, line 38: What if I want to see seconds even though `countDownData.hours >=1` or `countDownData.days >= 1`?  Maybe we can add a prop to `Countdown` as `alwaysShowSeconds`. This can be optional and default to `false`. 

```suggestion
  } else if (countDownData.hours >= 1) {
    items = [
      {id: "hours", count: countDownData.hours},
      {id: "minutes", count: countDownData.minutes},
      alwaysShowSeconds && {id: "seconds", count: countDownData.seconds}
    ].filter(Boolean) as CountDownItem[];
```

src/countdown/Countdown.tsx
```diff
@@ -0,0 +1,72 @@
...
+      items={items}>
+      {(item, itemTestId) => (
+        <ListItem customClassName={"countdown-box"}>
+          <p
+            data-testid={`${itemTestId}.${item.id}.count`}
+            className={"countdown-box__count"}>
+            {item.count}
+          </p>
+
+          <p className={"countdown-box__title"}>{item.id}</p>
```
bilgeyucel, line 65: Let's add `data-testid` here.
```suggestion
          <p className={"countdown-box__title"} data-testid={`${itemTestId}.${item.id}.title`}>{item.id}</p>
```

yasincaliskan: Thank you for your comments! 🎉

src/countdown/Countdown.tsx
```diff
@@ -0,0 +1,87 @@
...
+  startDate,
+  countDownIntervalInSeconds = 1,
+  alwaysShowSeconds = false,
+  onEnd,
+  customClassName
+}: CountdownProps) {
+  const countDownData = useCountDownTimer(startDate, countDownIntervalInSeconds);
+  let items: CountDownItem[] = [];
+
+  if (countDownData.days >= 1) {
```
mucahit, line 35: Let's make it a utility to increase the readability

`const items = generateCountDownItems(countDownData);`

mucahit (changes requested):

src/core/utils/hooks/useCountdownTimer.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+function useCountDownTimer(target: Date, cadence = MINUTE_IN_S): CountDownResults {
+  const interval = useRef<NodeJS.Timeout>();
+  const [countdown, setCountDown] = useState<CountDownResults>(
+    calculateRemainingTime(target)
+  );
+
+  useLayoutEffect(() => {
+    interval.current = setInterval(() => {
+      if (countdown.delta >= 0) {
```
edizcelik, line 22: Why are we comparing with the previous value of countdown ? We should calculate the new "remainingTime" data and compare with that
```suggestion
      const data = calculateRemainingTime(target);
      
      if (data.delta >= 0) {
          setCountDown(data);
```

src/core/utils/hooks/useCountdownTimer.ts
```diff
@@ -0,0 +1,41 @@
...
+import {calculateRemainingTime} from "../time/timeUtils";
+
+/**
+ * A React Hook that provides a countdown time
+ * @param {Date} target the target date we're counting down to
+ * @param {number} cadence the rate of the timer in seconds
+ *
+ * @returns {object} the CountDownResults object with ref
+ */
+function useCountDownTimer(target: Date, cadence = MINUTE_IN_S): CountDownResults {
```
edizcelik, line 14: Let's require the cadence in MS which would be more consistent with the rest of the JS tools, such as setInterval. You will need to update the JSDoc above as well
```suggestion
function useCountDownTimer(target: Date, cadence = MINUTE_IN_MS): CountDownResults {
```

src/core/utils/time/timeTypes.ts
```diff
@@ -0,0 +1,9 @@
+interface CountdownResults {
+  delta: number;
+  days: string | number;
+  hours: string | number;
+  minutes: string | number;
+  seconds: string | number;
```
edizcelik, line 6: These all must be numbers

src/core/utils/time/timeUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+    countdownResults.minutes = "00";
+  } else if (countdownResults.minutes <= maxDigit) {
+    countdownResults.minutes = `0${countdownResults.minutes}`;
+  }
+
+  if (countdownResults.seconds < 1) {
+    countdownResults.seconds = "00";
+  } else if (countdownResults.seconds <= maxDigit) {
+    countdownResults.seconds = `0${countdownResults.seconds}`;
+  }
```
edizcelik, line 45: These shouldn't be the responsibility of this function. This should return representation of the remaining time in MS to the target and breaks it down to days, hours, minutes, and seconds which have to be in numbers

src/core/utils/time/timeUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  return countdownResults;
+}
+
+function generateCountdownItems(
+  countdownData: CountdownResults,
+  secondVisibility: boolean
```
edizcelik, line 52: Let's swap the order and provide it as options
```suggestion
  {alwaysShowSeconds}: {alwaysShowSeconds: boolean}
  countdownData: CountdownResults
```

src/countdown/Countdown.tsx
```diff
@@ -0,0 +1,66 @@
...
+  customClassName
+}: CountdownProps) {
+  const countdownData = useCountdownTimer(startDate, countDownIntervalInSeconds);
+  let items = generateCountdownItems(countdownData, alwaysShowSeconds);
+
+  useLayoutEffect(() => {
+    if (countdownData.delta <= 0 && onEnd) {
+      onEnd();
+    }
+  }, [countdownData.delta, onEnd]);
```
edizcelik, line 40: We can't assume that the users of this component will remember to use `useCallback` when defining their `onEnd` handler. Let's make sure we save the callback and then use it within the effect

```
const savedOnEndCallback = useRef<CountdownProps["onEnd"]>();

useLayoutEffect(() => {
   savedOnEndCallback.current = onEnd;
});
```

Then use `savedOnEndCallback.current` within the layout effect that checks `countdownData.delta`

src/index.ts
```diff
@@ -43,6 +44,7 @@ export {
   FileUploadButton,
   Spinner,
   Tab,
+  Countdown,
```
edizcelik, line 47: Let's export `useCountdownTimer` hook as well in case the requirements are so different that they cannot use `Countdown` component

src/countdown/Countdown.tsx
```diff
@@ -0,0 +1,66 @@
...
+import React, {useLayoutEffect} from "react";
+import classNames from "classnames";
+
+import useCountdownTimer from "../core/utils/hooks/useCountdownTimer";
+import List from "../list/List";
+import ListItem from "../list/item/ListItem";
+import {generateCountdownItems} from "../core/utils/time/timeUtils";
+
+export interface CountdownProps {
+  testid: string;
```
edizcelik, line 12:
```suggestion
  testid?: string;
```

edizcelik (changes requested):

src/core/utils/hooks/useCountdownTimer.ts
```diff
@@ -0,0 +1,51 @@
...
+import {MINUTE_IN_MS} from "../time/timeConstants";
+import {CountdownResults} from "../time/timeTypes";
+import {calculateRemainingTime} from "../time/timeUtils";
+
+/**
+ * A React Hook that provides a countdown time
+ * @param {Date} target the target date we're counting down to
+ * @param {number} cadence the rate of the timer in milliseconds
+ *
+ * @returns {object} the CountDownResults object with ref
```
edizcelik, line 12:
```suggestion
 * @returns {object} the CountDownResults object
```

src/core/utils/hooks/useCountdownTimer.ts
```diff
@@ -0,0 +1,51 @@
...
+
+import {MINUTE_IN_MS} from "../time/timeConstants";
+import {CountdownResults} from "../time/timeTypes";
+import {calculateRemainingTime} from "../time/timeUtils";
+
+/**
+ * A React Hook that provides a countdown time
+ * @param {Date} target the target date we're counting down to
+ * @param {number} cadence the rate of the timer in milliseconds
+ *
```
edizcelik, line 11:
```suggestion
```

src/core/utils/hooks/useCountdownTimer.ts
```diff
@@ -0,0 +1,51 @@
+import {useEffect, useRef, useState, useLayoutEffect} from "react";
+
+import {MINUTE_IN_MS} from "../time/timeConstants";
+import {CountdownResults} from "../time/timeTypes";
+import {calculateRemainingTime} from "../time/timeUtils";
+
+/**
+ * A React Hook that provides a countdown time
+ * @param {Date} target the target date we're counting down to
+ * @param {number} cadence the rate of the timer in milliseconds
```
edizcelik, line 10:
```suggestion
 * @param {Date} target - the target date we're counting down to
 * @param {number} cadence - the rate of the timer in milliseconds
```

src/core/utils/hooks/useCountdownTimer.ts
```diff
@@ -0,0 +1,51 @@
...
+        setCountDown(data);
+      } else {
+        setCountDown({
+          delta: 0,
+          days: 0,
+          hours: 0,
+          minutes: 0,
+          seconds: 0
+        });
+      }
```
edizcelik, line 34: I don't know why you need to add this else block. Interval seems to be cleared at the other effect below when  `data.delta <= 0`, does it not ?

src/core/utils/time/timeUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+): CountdownItem[] {
+  const maxDigit = 9;
+  let items: CountdownItem[];
+
+  if (countdownData.days >= 1) {
+    items = [
+      {
+        id: "days",
+        count:
+          countdownData.days <= maxDigit ? `0${countdownData.days}` : countdownData.days
```
edizcelik, line 36: We can just use `padStart` method here instead of having to check against `maxDigit`
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/padStart
```suggestion
          countdownData.days.padStart(2, "0")
```

src/core/utils/time/timeUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+  return {
+    delta,
+    days: Math.floor(deltaInSeconds / DAY_IN_S),
+    hours: Math.floor((deltaInSeconds / HOUR_IN_S) % DAY_IN_HRS),
+    minutes: Math.floor((deltaInSeconds % HOUR_IN_S) / MINUTE_IN_S),
+    seconds: Math.floor((deltaInSeconds % HOUR_IN_S) % MINUTE_IN_S)
+  };
+}
+
+function generateCountdownItems(
```
edizcelik, line 24: This is a Countdown specific utility. It shouldn't be defined within `timeUtils`

src/countdown/Countdown.tsx
```diff
@@ -0,0 +1,71 @@
...
+  countDownIntervalInSeconds?: number;
+  alwaysShowSeconds?: boolean;
+  onEnd?: () => void;
+  customClassName?: string;
+}
+
+export interface CountdownItem {
+  id: "days" | "hours" | "minutes" | "seconds";
+  count: string | number;
+}
```
edizcelik, line 23: Let's create a `countdown.d.ts` and move these there

src/core/utils/time/timeUtils.ts
```diff
@@ -0,0 +1,96 @@
...
+  DAY_IN_HRS,
+  DAY_IN_S,
+  HOUR_IN_S,
+  MINUTE_IN_S,
+  SECOND_IN_MS
+} from "./timeConstants";
+import {CountdownResults} from "./timeTypes";
+import {CountdownItem} from "../../../countdown/Countdown";
+
+function calculateRemainingTime(target: Date): CountdownResults {
```
edizcelik, line 11: I'm not 100% happy about the name of the function and the type here. Maybe this is better. I think we should always try to add documentation for the utilities we define, so let's do that as well
```suggestion
function calculateRemainingTimeBreakdown(target: Date): RemainingTimeBreakdown {
```

edizcelik (changes requested):

src/countdown/Countdown.tsx
bilgeyucel: I wonder if we need `title` property for `CountdownItem`. Users may want to use other languages to display. Then, you will be changing line 63 with `item.title` @yasincaliskan

src/core/utils/hooks/useCountdownTimer.ts
yasincaliskan: Yes, looks we don't need. I thought we can guarantee the trigger `onEnd` function but `clearInterval` works well instead.

yasincaliskan: But I'm looking now, `clearInterval` is not working expectly. The following line doesn't execute and `countdown.delta` doesn't be zero. 
```
  if (data.delta >= 0) {
    setCountDown(data);
  }
```
So, in the `Countdown` component, the following line doesn't execute as well because `countdownData.delta` never is zero or negative value.
```
 useLayoutEffect(() => {
    if (countdownData.delta <= 0 && savedOnEndCallback.current) {
      savedOnEndCallback.current();
    }
  }, [countdownData.delta, onEnd]);
```
I guess, we should think something else.

src/countdown/Countdown.tsx
yasincaliskan: Actually, we can add an optional prop like;
```
title: {
  day: "days",
  hour: "hours",
  ...
}
```
If we need another language, we can override it. What do you think @edizcelik ?

src/core/utils/hooks/useCountdownTimer.ts
edizcelik: Weird but okay :+1:

src/countdown/Countdown.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+  useLayoutEffect(() => {
+    savedOnEndCallback.current = onEnd;
+  }, [onEnd]);
+
+  useLayoutEffect(() => {
+    if (countdownData.delta <= 0 && savedOnEndCallback.current) {
+      savedOnEndCallback.current();
+    }
+  }, [countdownData.delta, onEnd]);
```
edizcelik, line 32: It shouldn't have a dependency on `onEnd`

src/countdown/countdown.d.ts
```diff
@@ -0,0 +1,15 @@
...
+    alwaysShowSeconds?: boolean;
+    onEnd?: () => void;
+    customClassName?: string;
+  }
+
+  interface CountdownItem {
+    id: "days" | "hours" | "minutes" | "seconds";
+    count: string | number;
+  }
+}
```
edizcelik, line 15: Why are we using a namespace here ? Can't we just do the following ?
```suggestion
  interface CountdownProps {
    startDate: Date;
    testid?: string;
    countDownIntervalInSeconds?: number;
    alwaysShowSeconds?: boolean;
    onEnd?: () => void;
    customClassName?: string;
  }

  interface CountdownItem {
    id: "days" | "hours" | "minutes" | "seconds";
    count: string | number;
  }
```

src/countdown/util/countdownUtils.ts
```diff
@@ -0,0 +1,56 @@
+import {RemainingTimeBreakdown} from "../../core/utils/time/timeTypes";
+import {CountdownItem} from "../..";
```
edizcelik, line 2: Can you check if it works without any import ?
```suggestion
```

src/index.ts
```diff
@@ -24,7 +24,9 @@ import FileUploadButton, {
 } from "./button/file-upload/FileUploadButton";
 import Spinner, {SpinnerProps} from "./spinner/Spinner";
 import Tab, {TabItem, TabProps} from "./tab/Tab";
-import Textarea, { TextareaProps } from "./form/textarea/Textarea";
+import Countdown, {CountdownItem, CountdownProps} from "./countdown/Countdown";
```
edizcelik, line 27: These types aren't exported from Countdown file. Not sure whats going on here

src/countdown/util/countdownUtils.ts
yasincaliskan: Yes, looks working but TSC gave some warning about `Cannot find name X` for `CountdownItem` and `CountdownProps`.

src/countdown/util/countdownUtils.ts
```diff
@@ -0,0 +1,70 @@
...
+        count: countdownData.seconds.toString().padStart(2, "0"),
+        title: titleMap?.second || "seconds"
+      }
+    ];
+  }
+
+  if (alwaysShowSeconds && countdownData.hours >= 1) {
+    items.push({
+      id: "seconds",
+      count: countdownData.seconds,
```
edizcelik, line 62: Should we use padStart here too ?

src/countdown/Countdown.tsx
```diff
@@ -0,0 +1,59 @@
...
+          <p
+            data-testid={`${itemTestId}.${item.id}.count`}
+            className={"countdown-box__count"}>
+            {item.count}
+          </p>
+
+          <p
+            data-testid={`${itemTestId}.${item.id}.title`}
+            className={"countdown-box__title"}>
+            {item.id}
```
bilgeyucel, line 51: After adding a title
```suggestion
            {item.title}
```

bilgeyucel (changes requested):

edizcelik (approved):

bilgeyucel (approved):

Anlerkan (approved):

mucahit (approved):

jamcry (approved):

PR #78, opened 2021-04-19
feat(switch): Add Switch Component

yasincaliskan:
### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/77

image


src/switch/Switch.tsx
```diff
@@ -0,0 +1,41 @@
...
+      <input
+        className={"switch__controller"}
+        type="checkbox"
+        checked={isDisabled ? false : isToggledOn}
+        onChange={onToggle}
+      />
+
+      <span
+        className={classNames("switch__slider", {
+          "switch__slider--is-disabled": isDisabled
```
edizcelik, line 34: Let's add this to the container label

stories/10-Switch.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+        <StoryFragment>
+          <div style={{display: "flex", gap: "8px"}}>
+            <span>{"On"}</span>
+
+            <Switch
+              isToggledOn={state.switchState}
+              onToggle={() => setState({...state, switchState: !state.switchState})}
+            />
+
+            <span>{"Off"}</span>
```
edizcelik, line 26: Let's add a realistic example
```suggestion
            <span>{"Dark Theme"}</span>

            <Switch
              isToggledOn={state.switchState}
              onToggle={() => setState({...state, switchState: !state.switchState})}
            />
```

src/switch/_switch.scss
```diff
@@ -0,0 +1,70 @@
...
+
+  background-color: var(--switch-inactive-slider-bg);
+  border-radius: 34px;
+
+  cursor: pointer;
+
+  transition: 0.4s;
+
+  
+
```
bilgeyucel, line 53: No need this much space

stories/10-Switch.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+import {storiesOf} from "@storybook/react";
+
+import Switch from "../src/switch/Switch";
+import StateProvider from "./utils/StateProvider";
+import StoryFragment from "./utils/StoryFragment";
+
+storiesOf("Switch", module).add("Switch States", () => {
+  const initialState = {
+    switchState: true,
+    disabledState: false
```
bilgeyucel, line 11:
```suggestion
    isSwitchOn: true,
    isDisabled: false
```

src/switch/Switch.tsx
```diff
@@ -0,0 +1,40 @@
+import "./_switch.scss";
+
+import React from "react";
+import classNames from "classnames";
+
+export interface SwitchProps {
+  onToggle: () => void;
+  isToggledOn: boolean;
```
bilgeyucel, line 8: Can you use word `switch` instead of `toggle`?

yasincaliskan: Yes, we can use it. But as typical usage, it similar to `Toggle` component behavior. Actually, I think `onToggle` more descriptive then. We can take the team ideas as well

src/switch/_switch.scss
```diff
@@ -0,0 +1,68 @@
...
+  bottom: 0;
+  left: 0;
+  right: 0;
+
+  background-color: var(--switch-inactive-slider-bg);
+  border-radius: 34px;
+
+  cursor: pointer;
+
+  transition: 0.4s;
```
jamcry, line 50: Do we have a constant `--default-animation-duration` in this project? If not, it may be a good idea to define it and use it here (as far as i can see we also use animation in several places such as radio input and dropdown list item hover state)
This way all animations will be consistent, and package users can configure this value

stories/10-Switch.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+  const initialState = {
+    isSwitchOn: true,
+    isDisabled: false
+  };
+
+  const darkTheme = (
+    <style>{`
+    .sb-show-main {
+      background-color: gray;
+    }
```
jamcry, line 18:
```suggestion
      background-color: gray;
      color: white;
    }
```
Not a big deal, but i think it is a good touch 😄

src/switch/_switch.scss
```diff
@@ -0,0 +1,68 @@
...
+
+  display: inline-block;
+
+  width: 60px;
+  height: 34px;
+
+  &--is-disabled {
+    cursor: not-allowed;
+    pointer-events: none;
+  }
```
jamcry, line 16: `    cursor: not-allowed;`
This is not working for some reason, can you check it please? See the screenshots below:
When i hover disabled switch, cursor is not changed

Disabled switch:
Screen Shot 2021-04-23 at 12 37 16 PM

Disabled button:
Screen Shot 2021-04-23 at 12 37 27 PM

src/switch/Switch.tsx
jamcry: I also think `toggle` is OK here, but as an alternative, we can the name convention of checkbox directly:
```ts
isChecked
onChange
```
Maybe this would be a more predictable name for users

jamcry (changes requested):

src/switch/_switch.scss
yasincaliskan: Yes, we already have. I didn't see that. I've changed with `$animation-duration`. Thank you 🎉

src/switch/_switch.scss
yasincaliskan: When I passed `pointer-event: none;` property to `.switch__slider` for the disabled state, it works! 🚀

src/switch/Switch.tsx
yasincaliskan: Yeah, we can but `isChecked` looks better than `isToggledOn` for the checkbox element from the name consistency point.

jamcry (approved):

Anlerkan (approved):

edizcelik (approved):

src/switch/Switch.tsx
edizcelik: I think `onToggle` is fine because we are toggling the switch on and off whenever that function is executed. Maybe we can change `isToggledOn` as just `isOn` but they all look fine to me

mucahit (approved):

bilgeyucel (approved):

PR #79, opened 2021-04-19
feat(toggle): Add Toggle Component

yasincaliskan:
### Description
- Implementation `Compound Component` pattern.
- `Toggle` should have `onToggle` function and children nodes like `Toggle.Item`.

### Usage
```
  <Toggle isMultiple={true} onToggle={handleToggle}>
      <Toggle.Item id="react">{"React"}</Toggle.Item>
      <Toggle.Item id="vue">{"Vue"}</Toggle.Item>
      <Toggle.Item id="angular">{"Angular"}</Toggle.Item>
  </Toggle>
```

### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/74

image

image


src/toggle/Toggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+        {children}
+      </ToggleContext.Provider>
+    </div>
+  );
+
+  function onToggleItem(id: string) {
+    onToggle(id);
+
+    return id;
+  }
```
jamcry, line 68: Why does this function returns `id`? I couldn't figure out the use case here

src/toggle/item/_toggle-item.scss
```diff
@@ -0,0 +1,35 @@
...
+
+  &--is-selected {
+    color: var(--toggle-active-color);
+    background-color: var(--toggle-active-bg);
+  }
+
+  &--is-disabled {
+    color: var(--toggle-disabled-color);
+    background-color: var(--toggle-disabled-bg);
+
```
jamcry, line 22: I think this would be a good addition
```suggestion
cursor: not-allowed;
```

src/toggle/Toggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+    "toggle--is-horizontal": position === "horizontal",
+    "toggle--is-vertical": position === "vertical",
+    "toggle--is-disabled": isDisabled
+  });
+  const [selectedToggleItemsState, setSelectedToggleItemsState] = useState<
+    ToggleItemShape[]
+  >([]);
+
+  return (
+    <div data-testid={testid} className={toggleClassName}>
```
jamcry, line 51: I believe we should wrap Toggle component body with a `List` instead of div, because it is actually a list with `li` items (`ToggleItem`)

jamcry (commented):
I have a question, this toggle functionality is very similar to Tab component functionality. I am not very familiar with the implementations, but do you think would using `Toggle` for `Tab`'s header part make it simpler? Not critical now, but maybe we can create an improvement issue

src/toggle/Toggle.tsx
yasincaliskan: I thought we could access the item `id` that was toggled. `onToggleItem` function handle both of `ToggleItem` and the parent components handle function. But if it's confusing, we can think of an easier usage 🤔 

```
 onToggleItem: (id: string) => id
```

src/toggle/Toggle.tsx
yasincaliskan: `ToggleItem` is a `ListItem` variant actually but `Toggle` provides a context. We don't know how many items should exist in `ToggleContext` and we can't pass `items` prop that `ListItem's`. I'm not sure we can use `List`.

yasincaliskan:
Yes, good point. They are similar to each other. But I think Tab's functionality is just to switch the tab and tab's content. However Toggle can have any children and have more functionality. 
Eventually,  I think Toggle can include Tab component.

src/toggle/Toggle.tsx
jamcry: Yeah I didn't think of that. Then, I think we can directly use `ul` here (instead of `div`)
Because at any case, the children will be `li`s

src/toggle/Toggle.tsx
jamcry: But we already pass `id` to `onToggle`, so the parent will know the `id`. Am I missing something ? 🤔

src/toggle/Toggle.tsx
yasincaliskan: Absolutely right! Thank you 🚀

src/toggle/Toggle.tsx
yasincaliskan: While the context creating, `onToggleItem` takes an `id` parameter and returns this. Because `Toggle` should know clicked item for getting action. In the don't return `id` case, we can't handle specific `onToggle` action.
If you know another easier way, please share with me :)

jamcry: I see now, thanks 👍

src/toggle/Toggle.tsx
```diff
@@ -0,0 +1,73 @@
...
+    "toggle--is-horizontal": position === "horizontal",
+    "toggle--is-vertical": position === "vertical",
+    "toggle--is-disabled": isDisabled
+  });
+  const [selectedToggleItemsState, setSelectedToggleItemsState] = useState<
+    ToggleItemShape[]
+  >([]);
+
+  return (
+    <li data-testid={testid} className={toggleClassName}>
```
jamcry, line 51: I think this should be `ul`, not `li`

jamcry (approved):

edizcelik (changes requested):

src/toggle/item/ToggleItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+      clickableListItemProps={{onClick: handleToggle}}>
+      {children}
+    </ListItem>
+  );
+
+  function handleToggle() {
+    if (isMultiple) {
+      if (isSelected) {
+        setSelectedToggleItemsState([
+          ...selectedToggleItemsState.filter((item) => item.id !== id)
```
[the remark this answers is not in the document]
yasincaliskan: If select an already selected item, the item should remove from the selected array. Is it the wrong approach? 🤔

stories/11-Toggle.stories.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+import {Toggle} from "../src/toggle/Toggle";
+
+storiesOf("Toggle", module)
+  .add("Toggle States", () => {
+    return (
+      <div style={{width: "500px"}}>
+        <span>{"Switch Toggle - 2 Options"}</span>
+        <Toggle onToggle={(e) => console.log(e)}>
+          <Toggle.Item id={"on"}>{"On"}</Toggle.Item>
```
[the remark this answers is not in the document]
yasincaliskan: Yes, we can change it. But `data-id` might not proper name because of `dash`. However, we can use `dataId`.

src/toggle/item/ToggleItem.tsx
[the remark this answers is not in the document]
edizcelik: Thats correct but we don't need to create a new array from the newly created filtered array

src/toggle/Toggle.tsx
```diff
@@ -0,0 +1,74 @@
...
+  isDisabled,
+  customClassName
+}: ToggleProps) {
+  const toggleClassName = classNames("toggle", customClassName, {
+    "toggle--is-horizontal": position === "horizontal",
+    "toggle--is-vertical": position === "vertical",
+    "toggle--is-disabled": isDisabled
+  });
+  const [selectedToggleItemsState, setSelectedToggleItemsState] = useState<string[]>(
+    selectedItems
```
edizcelik, line 34: this should be `initialSelectedItems`. IF they provide `selectedItems`, then we should pass `selectedItems` directly to the context instead of using `selectedToggleItemsState`. Please add an example to the Storybook that shows the use of `selectedItems` – you will see that your current implementation is not working properly in that case.

src/toggle/item/ToggleItem.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+function ToggleItem({
+  children,
+  dataId,
+  customClassName,
+  isDisabled,
+  testid
+}: ToggleItemProps) {
+  const {selectedToggleItemsState, onToggleItem} = useToggle();
+  const isSelected = selectedToggleItemsState.some((item) => item === dataId);
```
edizcelik, line 25:
```suggestion
  const isSelected = selectedToggleItemsState.includes(dataId);
```

src/toggle/item/_toggle-item.scss
```diff
@@ -0,0 +1,38 @@
...
+    cursor: not-allowed;
+    
+    .list-item__click-wrapper {
+      pointer-events: none;
+    }
```
edizcelik, line 27: With the latest changes, there wouldn't be this click wrapper div when isDisabled is true, would there ?

src/toggle/util/ToggleContext.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+type ToggleContextValue = {
+  selectedToggleItemsState: string[];
+  onToggleItem: (dataId: string) => void;
+};
+
+const ToggleContext = createContext<ToggleContextValue>({
+  selectedToggleItemsState: [] as string[],
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  onToggleItem: () => {}
```
edizcelik, line 11: You shouldn't disable this Eslint rule but fix it so that ESLint would be happy

src/toggle/util/ToggleContext.tsx
```diff
@@ -0,0 +1,24 @@
...
+});
+
```
edizcelik, line 13: Let's add a displayName for the ToggleContext
```suggestion
ToggleContext.displayName = "ToggleContext";
```

edizcelik (changes requested):

stories/11-Toggle.stories.tsx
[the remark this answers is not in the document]
edizcelik: Ok let's have it as `dataId` then

src/toggle/item/_toggle-item.scss
yasincaliskan: Right, but we still need a child selector. Because the following hierarchy should be for display disabled cursor;
```.scss
.parent-selector {
  cursor: not-allowed;

    * {
      pointer-events: none;
    }
}
```
Can we use like that?

src/toggle/Toggle.tsx
```diff
@@ -0,0 +1,48 @@
...
+
+import React from "react";
+import classNames from "classnames";
+
+import ToggleItem from "./item/ToggleItem";
+import {ToggleContext} from "./util/ToggleContext";
+
+export interface ToggleProps {
+  children: React.ReactNode;
+  onToggle: (dataId: string) => void;
```
edizcelik, line 11: Not sure about this but I think we should handle this within Toggle and pass the newly generated "selectedItems" array to the parent. What do you think ?
```suggestion
  onChange: (selectedItems: string[]) => void;
```

This way we can move the following function to Toggle and handle the manipulation within Toggle component itself:

```
function handleToggle(dataId: string, state: any) {
  let selectedItems;

  if (state.includes(dataId)) {
    selectedItems = state.filter((item) => item !== dataId);
  } else {
    selectedItems = [...state, dataId];
  }

  return selectedItems;
}
```

cc: @bilgeyucel @mucahit @jamcry @Anlerkan

yasincaliskan: In this case, we need `isMultiple` prop. Because the user might want to select only one item. If `isMultiple` is false, we can return just one item.

jamcry: I think Ediz's suggestion makes sense. Currently, we are already keeping array for each state
```
  switch: [],
  devices: [],
  frameworks: []
```
We can add something like `canSelectMultiple`. I think it was named like that in `TypeaheadSelect`.

jamcry (approved):

bilgeyucel (approved):

edizcelik (approved):

Anlerkan (approved):

src/toggle/_toggle.scss
```diff
@@ -0,0 +1,30 @@
+.toggle {
+  display: grid;
+  
+  overflow: hidden;
+  
+  padding: 0;
+
+  border: 1px solid gray;
```
mucahit, line 8: Let's get the color from toggle specific css variable
```css
--toggle-border-color: gray;
```

mucahit (approved):

Issue #75, opened 2021-04-19
Allow custom Spinner component for Button


edizcelik:

mucahit:
Also, we can replace the internal Spinner with this one 
TSX: https://gist.github.com/mucahit/5c381504558ca783a5135850ec1cd619
SCSS: https://gist.github.com/mucahit/cd206fa1d18b75a95ab125e766a6b467

PR #80, opened 2021-04-20
feat(button): Allow custom Spinner for Button

yasincaliskan:
### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/75

image


src/button/Button.tsx
```diff
@@ -54,6 +56,14 @@ function Button({
     "button--is-inactive": isButtonDisabled,
     "button--is-pending": shouldDisplaySpinner
   });
+  const displaySpinner = customSpinner || (
```
mucahit, line 59:
```suggestion
  const spinnerContent = customSpinner || (
```

src/form/input/file/FileInput.tsx
```diff
@@ -38,6 +40,9 @@ function FileInput({
   const labelClassName = classNames("file-input__label", customLabelClassName, {
     "file-input__label--is-disabled": isInputDisabled
   });
+  const displaySpinner = customSpinner || (
```
mucahit, line 43: Same here

jamcry (approved):

src/button/Button.tsx
```diff
@@ -54,6 +56,14 @@ function Button({
     "button--is-inactive": isButtonDisabled,
     "button--is-pending": shouldDisplaySpinner
   });
+  const spinnerContent = customSpinner || (
+    <Spinner
+      customClassName={classNames("button__spinner", {
+        [`${customClassName}__spinner`]: customClassName
```
edizcelik, line 62: Let's not do this because it may lead to weird classes. For example, if we provide customClassName as `button--small button--primary` then this results in adding `button--small button--primary__spinner`. Let's just use `button__spinner` when we want to override the styles of the default Spinner

src/toggle/util/toggleHooks.ts
```diff
@@ -0,0 +1,12 @@
...
+import {ToggleContext} from "../Toggle";
+
+/**
+ * @returns {Object} Current value of ToggleContext
+ */
+function useToggleContext() {
+  return useContext(ToggleContext);
+}
+
+export {useToggleContext};
```
edizcelik, line 12: Why is this included here ? 🤔

yasincaliskan: Ops! It's really weird. I guess my local branches are confused.

src/button/Button.tsx
yasincaliskan: Actually, this [PR](https://github.com/Hipo/react-ui-toolkit/pull/81) includes this solution. If it's OK, I can close this PR.

mucahit (approved):

Anlerkan (approved):

bilgeyucel (approved):

edizcelik (approved):

PR #81, opened 2021-04-20
refactor(spinner): Simplify Spinner Component

yasincaliskan:
### Note
- Spinner takes just color as a prop.

### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/75


src/button/Button.tsx
```diff
@@ -74,12 +74,7 @@ function Button({
       disabled={isButtonDisabled}
       aria-label={ariaLabel}>
       {shouldDisplaySpinner ? (
-        <Spinner
-          customClassName={classNames("button__spinner", {
-            [`${customClassName}__spinner`]: customClassName
-          })}
-          aria-label={"Button spinner visible. Button inactivated."}
-        />
+        <Spinner aria-label={"Button spinner visible. Button inactivated."} />
```
mucahit, line 77: Should we keep `button__spinner` className?

yasincaliskan: I think it seems we don't need it. But we can add `customClassName` prop again and we can pass that className.

mucahit: Yeah, let's add `customClassName` and pass that. We definitely gonna need `customClassName` because we're using the Spinner outside of the toolkit e.g. `RouteLoading`

jamcry (approved):

Anlerkan (approved):

src/dropdown/Dropdown.tsx
```diff
@@ -149,11 +149,7 @@ function Dropdown<OptionIdShape extends string>({
         {selectedOption ? selectedOption.title : deselectOptionTitle || placeholder}
       </span>
 
-      {areOptionsFetching ? (
-        <Spinner spinnerColor={"black"} backgroundColor={"#EBEBEB"} />
-      ) : (
-        <CaretDownIcon aria-hidden={true} />
-      )}
+      {areOptionsFetching ? <Spinner /> : <CaretDownIcon aria-hidden={true} />}
```
edizcelik, line 152: Let's pass "dropdown__spinner" as a customClassName to Spinner. Also can we allow passing a customSpinner to Dropdown as well?

src/form/input/file/FileInput.tsx
```diff
@@ -60,7 +60,7 @@ function FileInput({
         data-testid={`${testid}.label`}>
         {children}
 
-        {isPending && <Spinner customClassName={"file-input__spinner"} />}
```
edizcelik, line 63: Let's keep the classname and allow passing a customSpinner to FileInput

src/select/typeahead/TypeaheadSelect.tsx
```diff
@@ -134,11 +134,7 @@ function TypeaheadSelect({
           onQueryChange={handleKeywordChange}
           onKeyDown={handleKeyDown}
           rightIcon={
-            areOptionsFetching ? (
-              <Spinner spinnerColor={"#EBEBEB"} backgroundColor={"white"} />
-            ) : (
-              <CaretDownIcon aria-hidden={true} />
-            )
+            areOptionsFetching ? <Spinner /> : <CaretDownIcon aria-hidden={true} />
```
edizcelik, line 137: Same here

src/dropdown/Dropdown.tsx
yasincaliskan: Why not? I think we can provide this with a `customSpinner` prop.

src/spinner/_spinner.scss
```diff
@@ -1,45 +1,16 @@
...
-
-  width: 75%;
-  height: 75%;
-
-  margin: auto;
-
-  border-radius: 50%;
-
-  content: "";
+  animation-name: spinner;
```
mucahit, line 2: Let's make it shorthand

```suggestion
  animation: spinner 1.7s linear infinite;
```

mucahit (approved):

bilgeyucel (approved):
🚀

edizcelik (approved):

Issue #76, opened 2021-04-19
Add TimeInput component


edizcelik:
Our current component is just an Input that handles various input formats and wisely determines what the time would be from the entered value. We should combine that with a dropdown off available options, similar to Google Calendar's time select component.

PR #83, opened 2021-04-22
feat(time-input): Add TimeInput Component

yasincaliskan:
### Description
- Implementation `TimeInput` component
- Install  `date-fns` and `date-fns-tz` packages

### Notes
- `TimeInput` can take `selectedDate` and `startTime` props as optional. 

### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/76


src/core/utils/number/numberUtils.ts
```diff
@@ -0,0 +1,25 @@
...
+    return finalNumber;
+  };
+}
+
+/**
+ * @param {number} x number to check
+ * @return {boolean} Whether x is an integer or not
+ */
+function isInteger(x: number): boolean {
+  return String(x).split(".")[1] === undefined;
```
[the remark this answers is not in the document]
yasincaliskan: Looks better, thank you 🎉

edizcelik (commented):
Just to provide the most flexibility I think we should have three components exported for this:

- TimeInput  <- implements only the input part
- TimeDropdown <- implements the dropdown
- TimeSelect <- Uses TimeInput and TimeDropdown and combines them to offer full TimeSelect functionality

package.json
```diff
@@ -42,6 +42,8 @@
     "@typescript-eslint/parser": "3.2.0",
     "babel-loader": "8.1.0",
     "css-loader": "3.5.3",
+    "date-fns": "2.21.1",
+    "date-fns-tz": "1.1.4",
```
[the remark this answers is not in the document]
yasincaliskan: Exactly should be in dependencies. Thank you!

src/form/time/TimeInput.tsx
```diff
@@ -0,0 +1,119 @@
...
+      startDate: selectedDate
+    })
+  );
+  const [value, setValue] = useState({
+    input: timeStringOfDate,
+    dropdown: null
+  });
+
+  useEffect(() => {
+    setDropdownOptions(generateTimeInputDropdownOptions({startTime: formattedStartTime}));
```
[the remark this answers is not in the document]
yasincaliskan: In the current version that separated Time components, needs to only `startTimeString`. I think we should need only one of `startDate` or `startTimeString`. `startDate: Date;` can better convention I think. What do you think?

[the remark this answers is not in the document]
edizcelik: I think we should be providing both of them otherwise wouldn't the constructed Date object for each option be wrong ?

src/form/time/dropdown/util/timeDropdownUtils.ts
```diff
@@ -0,0 +1,111 @@
...
+    title: timeString,
+    subtitle: generateTimeDropdownOptionSubtitle(differenceInMinutes),
+    context: {
+      date,
+      differenceInMinutes
+    }
+  };
+}
+
+function generateTimeDropdownDates(
```
edizcelik, line 39:
```suggestion
function generateTimeDropdownOptionsFromStartTimeDate(
```

src/form/time/dropdown/util/timeDropdownUtils.ts
```diff
@@ -0,0 +1,111 @@
...
+    const startTimeDate = parse(
+      formatTimeStringTo12hFormatWithMeridiem(startTime),
+      DATE_FORMAT.LONG_TIME_FORMAT,
+      startDate || new Date()
+    );
+
+    // First four options are separated by only 15 mins
+    const firstHourOptions = generateTimeDropdownDates(startTimeDate, {
+      // eslint-disable-next-line no-magic-numbers
+      arrayLength: 4,
```
edizcelik, line 74:
```suggestion
      arrayLength: Math.floor(HOUR_IN_MINUTES / TIME_DROPDOWN_OPTION_INTERVALS.FIRST_HOUR),
```

src/form/time/dropdown/TimeDropdown.tsx
```diff
@@ -0,0 +1,79 @@
...
+import {generateTimeDropdownOptions} from "./util/timeDropdownUtils";
+
+export type TimeDropdownOption = DropdownOption<
+  string,
+  {date: Date; differenceInMinutes: number}
+>;
+
+export interface TimeDropdownProps {
+  testid: string;
+  startTimeString: string;
```
edizcelik, line 17: I think we need to optionally expect a `startDate`  as well. Otherwise, TimeDropdown will always have the options within today, https://github.com/Hipo/react-ui-toolkit/pull/83/files#diff-8852021072415203fa35795a478d99809c89a9f47c3ed888013a275b44916bbaR65.

src/form/time/select/TimeSelect.tsx
```diff
@@ -0,0 +1,102 @@
...
+import {formatDateWithOptions} from "../../../core/utils/time/timeUtils";
+import {DATE_FORMAT} from "../../../core/utils/time/timeConstants";
+import {DropdownOption} from "../../../dropdown/list/item/DropdownListItem";
+import TimeInput from "../input/TimeInput";
+import TimeDropdown from "../dropdown/TimeDropdown";
+
+export type TimeSelectDropdownOption = DropdownOption<
+  string,
+  {date: Date; differenceInMinutes: number}
+>;
```
edizcelik, line 15:
```suggestion
export type TimeSelectDropdownOption = TimeDropdownOption;
```

Or, we can remove `TimeSelectDropdownOption` because I think they can just use `TimeDropdownOption`

src/form/time/select/TimeSelect.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export interface TimeSelectProps {
+  onChange: (timeString: string) => void;
+  startTime: Date;
+  testid: string;
```
edizcelik, line 20: We are keeping the testids optional for now

src/form/time/select/TimeSelect.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const timeStringOfDate = selectedDate
+    ? formatDateWithOptions({
+        format: DATE_FORMAT.LONG_TIME_FORMAT,
+        shouldShiftDateToCompensateForTimezone: false
+      })(selectedDate)
+    : "";
+  const formattedStartTime = formatDateWithOptions({
+    format: DATE_FORMAT.LONG_TIME_FORMAT,
+    shouldShiftDateToCompensateForTimezone: false
+  })(startTime);
```
edizcelik, line 56:
```suggestion
    ? timeFormatter(selectedDate)
    : "";
  const formattedStartTime = timeFormatter(startTime);
```

src/form/time/select/TimeSelect.tsx
```diff
@@ -0,0 +1,102 @@
...
+export type TimeSelectDropdownOption = DropdownOption<
+  string,
+  {date: Date; differenceInMinutes: number}
+>;
+
+export interface TimeSelectProps {
+  onChange: (timeString: string) => void;
+  startTime: Date;
+  testid: string;
+  selectedDate?: Date | null;
```
edizcelik, line 21: I guess this is `startDate`

src/form/time/input/TimeInput.tsx
```diff
@@ -0,0 +1,67 @@
+import React, {useState} from "react";
+import classNames from "classnames";
+
+import Input from "../../input/Input";
+import {DATE_FORMAT} from "../../../core/utils/time/timeConstants";
+import {formatDateWithOptions, parseTime} from "../../../core/utils/time/timeUtils";
+
+export interface TimeInputProps {
+  testid: string;
+  onChange: (timeString: string) => void;
```
edizcelik, line 10: This doesn't get a value prop. How can its state be controlled by the parent ?

src/form/time/input/TimeInput.tsx
```diff
@@ -0,0 +1,67 @@
...
+  customClassName,
+  hasError
+}: TimeInputProps) {
+  const timeStringOfDate = selectedDate
+    ? formatDateWithOptions({
+        format: DATE_FORMAT.LONG_TIME_FORMAT,
+        shouldShiftDateToCompensateForTimezone: false
+      })(selectedDate)
+    : "";
+  const [value, setValue] = useState(timeStringOfDate);
```
edizcelik, line 37: I think it should also get a `value` prop to turn this component into a controlled component. When the value is passed from the parent, TimeInput should delegate the handling of its state to its parent. Also, it seems we want to calculate `timeStringOfDate` just once.
```suggestion
  const [value, setValue] = useState(() => selectedDate
    ? formatDateWithOptions({
        format: DATE_FORMAT.LONG_TIME_FORMAT,
        shouldShiftDateToCompensateForTimezone: false
      })(selectedDate)
    : "");
```

src/form/time/select/TimeSelect.tsx
```diff
@@ -0,0 +1,102 @@
...
+    format: DATE_FORMAT.LONG_TIME_FORMAT,
+    shouldShiftDateToCompensateForTimezone: false
+  })(startTime);
+  const [value, setValue] = useState({
+    input: timeStringOfDate,
+    dropdown: null
+  });
+
+  const timeDropdownHeader = (
+    <TimeInput
```
edizcelik, line 63: For example, here it seems we want to turn TimeInput into a controlled component and provide the `value` prop

src/form/time/select/TimeSelect.tsx
```diff
@@ -0,0 +1,102 @@
...
+      })(selectedDate)
+    : "";
+  const formattedStartTime = formatDateWithOptions({
+    format: DATE_FORMAT.LONG_TIME_FORMAT,
+    shouldShiftDateToCompensateForTimezone: false
+  })(startTime);
+  const [value, setValue] = useState({
+    input: timeStringOfDate,
+    dropdown: null
+  });
```
edizcelik, line 60: Let's have two separate useStates for these

src/form/time/select/TimeSelect.tsx
```diff
@@ -0,0 +1,102 @@
...
+    setValue({...value, input: timeString});
+
+    onChange(value.input);
+  }
+
+  function handleSelect(option: DropdownOption | null) {
+    setValue({
+      input: option!.title,
+      dropdown: option!.context
+    });
```
edizcelik, line 98: I believe we would want to call onChange here too

src/form/time/select/TimeSelect.tsx
```diff
@@ -0,0 +1,102 @@
...
+import TimeInput from "../input/TimeInput";
+import TimeDropdown from "../dropdown/TimeDropdown";
+
+export type TimeSelectDropdownOption = DropdownOption<
+  string,
+  {date: Date; differenceInMinutes: number}
+>;
+
+export interface TimeSelectProps {
+  onChange: (timeString: string) => void;
```
edizcelik, line 18: I think we would want to pass a "selectedTime" prop instead of `startTime` and `startDate`. This way the props make more sense and it is obvious that TimeSelect is actually a controlled component.

stories/0-Form.stories.tsx
```diff
@@ -208,6 +215,115 @@ storiesOf("Form", module)
       </StateProvider>
     </Fragment>
   ))
+  .add("Time Input", () => {
```
edizcelik, line 218: Let's create a separate story these. You can name the story as `Time Select Fields`. The file name can be similar to this one: `11-TimeSelectFields.stories.tsx`

stories/0-Form.stories.tsx
```diff
@@ -208,6 +215,115 @@ storiesOf("Form", module)
...
+    const lessonEndTime = new Date(new Date().setMinutes(new Date().getMinutes() + 30));
+
+    return (
+      <Fragment>
+        <FormField label={"Appointment Time - TimeInput"}>
+          <TimeInput
+            testid={"appointment-time"}
+            selectedDate={new Date()}
```
edizcelik, line 226:
```suggestion
            initialDateTime={new Date()}
```

edizcelik (commented):
It seems we are creating some state duplications between parents and individual Time fields. Most of this is because they cannot be controlled by the parent. It might be better to be able to delegate the state to the parent for form fields. Let's rethink the overall structure and how the components play with each other, and try to improve.

src/form/time/select/TimeSelect.tsx
yasincaliskan: Right, we don't need duplicated types. I'll remove `TimeSelectDropdownOption`. Thanks!

package.json
```diff
@@ -97,7 +97,8 @@
   },
   "dependencies": {
     "classnames": "2.3.1",
-    "react-textarea-autosize": "8.3.2"
```
mucahit, line 100: Why did we remove `react-textarea-autosize`?

src/form/time/dropdown/TimeDropdown.tsx
```diff
@@ -0,0 +1,81 @@
...
+  startDate?: Date;
+  startTimeString?: string;
+  customHeader?: React.ReactNode;
+  icon?: React.ReactNode;
+  hasDeselectOption?: boolean;
+  isDisabled?: boolean;
+  customClassName?: string;
+}
+
+function TimeDropdown({
```
mucahit, line 28: Do we really need this component, it seems it doesn't add much customization. We can still provide the necessary utilities to build this component and it'll be enough. (`generateTimeDropdownOptions`)

src/core/utils/time/timeUtils.ts
```diff
@@ -20,4 +28,173 @@ function calculateRemainingTimeBreakdown(target: Date): RemainingTimeBreakdown {
   };
 }
 
-export {calculateRemainingTimeBreakdown};
+/**
```
mucahit, line 31: Added line breaks between paragraphs to improve readability

```suggestion
/**
 * Formats a Date object into a human friendly string
 *
 * Note on usage:
 *    Whenever the timezone information is not available and a date object that is passed to this 
 *    function is generated from a ISO-8601 date string (yyyy-MM-dd), `shouldShiftDateToCompensateForTimezone` option should be set to `true`.
 * 
 *    Otherwise, displayed date would be inconsistent with the data. For example, for a user with Eastern Daylight Time (to generalize, any user that has a negative UTC offset), 
 *    the following Date object is created for "2007-05-16": Tue May 15 2007 20:00:00 GMT-0400 (Eastern Daylight Time) {}. Therefore, 15 May appears on the screen. 
 *    By passing `new Date("2007-05-16")` value to `compensateForTimezone` utility, we fix this problem.
 * 
 *    When the timezone information is passed to `formatDateWithOptions`, this extra compensation is redundant as `date-fns-tz/utcToZonedTime` handles it correctly.
 *
 * @param {object} options FormatDateUtilOptions
 * @return {string} Formatted date
 */
```

src/core/utils/time/timeUtils.ts
```diff
@@ -20,4 +28,173 @@ function calculateRemainingTimeBreakdown(target: Date): RemainingTimeBreakdown {
...
+    return timeZone
+      ? timezonedFormat(dateToFormat, format, {
+          locale: enCA,
+          timeZone
+        })
+      : formatWithOptions({locale: enCA}, format)(dateToFormat);
+  };
+}
+
+/**
```
mucahit, line 68:
```suggestion
/**
 * Shifts the time of the Date object by the system's UTC offset to avoid timezone issues.
 *
 * For example, for a user with Eastern Daylight Time (to generalize, any user that has a negative UTC offset), 
 * the following Date object is created for "2007-05-16": Tue May 15 2007 20:00:00 GMT-0400 (Eastern Daylight Time) {}.
 * 
 * By using this utility we get the following Date object: `compansateForTimezone(new Date("2007-05-16"))` = Tue May 16 2007 00:00:00 GMT-0400 (Eastern Daylight Time) {}
 *
 * @param {Date} date Date to shift
 * @return {Date} shifted Date
 */
```

src/form/time/dropdown/util/timeDropdownUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+    .map<TimeDropdownOption>((_item, index) =>
+      generateTimeDropdownOptionFromStartTimeDate(
+        startTimeDate,
+        startTimeOffsetInMins + index * interval
+      )
+    );
+}
+
+function generateTimeDropdownOptions(options?: {
+  startDate?: Date | null;
```
mucahit, line 54: Can we do it with a single `startTime` prop? `startTime?: Date | string | null`

src/form/time/dropdown/util/timeDropdownUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (startTime) {
+    const startTimeDate = parse(
+      formatTimeStringTo12hFormatWithMeridiem(startTime),
+      DATE_FORMAT.LONG_TIME_FORMAT,
+      startDate || new Date()
+    );
+
+    // First four options are separated by only 15 mins
+    const firstHourOptions = generateTimeDropdownOptionsFromStartTimeDate(startTimeDate, {
+      // eslint-disable-next-line no-magic-numbers
```
mucahit, line 69: No need to disable it

src/form/time/dropdown/util/timeDropdownUtils.ts
```diff
@@ -0,0 +1,107 @@
...
+      arrayLength: Math.floor(HOUR_IN_MINS / TIME_DROPDOWN_OPTION_INTERVALS.FIRST_HOUR),
```
mucahit, line 70: It would be great if we can get the `TIME_DROPDOWN_OPTION_INTERVALS` object dynamically from the params

package.json
yasincaliskan: It should be a mistake. Seems like our `Textarea` component is still using it. I added again.

src/core/utils/time/timeUtils.ts
yasincaliskan: Looks great. Thanks!

src/form/time/dropdown/TimeDropdown.tsx
yasincaliskan: Yes, we don't need `TimeDropdown` and `TimeSelect` components for now.

src/form/time/select/TimeSelect.tsx
yasincaliskan: We removed `TimeSelect` and `TimeDropdown` component as we discussed.

src/form/time-input/TimeInput.tsx
```diff
@@ -0,0 +1,65 @@
+import React from "react";
+import classNames from "classnames";
+
+import Input from "../input/Input";
+import {DATE_FORMAT} from "../../core/utils/time/timeConstants";
+import {formatDateWithOptions, parseTime} from "../../core/utils/time/timeUtils";
+
+export interface TimeInputProps {
```
Anlerkan, line 8: Looks great. Thanks! 💯 

I have a few questions/requests about `TimeInputProps`
- I think we should get native input HTML attributes also. 
-  `testid` should be optional like we did on InputProps.
-  Should we replace `icon` prop to `leftIcon` and `rightIcon` props? I think it can be more flexible this way.
- In short, I think we can generate `TimeInputProps` from `InputProps` like this:
```tsx
export type TimeInputProps = Omit<
  InputProps,
  "localizationOptions" | "onChange" | "value"
> & {
  initialDateTime?: Date | null;
  onChange: (timeString: string) => void;
  value: string;
};
```

Here is the `InputProps` for reference:
```tsx
export type InputProps = Omit<
  React.InputHTMLAttributes<HTMLInputElement>,
  "disabled" | "name" | "className"
> & {
  name: string;
  type?: InputTypes;
  testid?: string;
  leftIcon?: React.ReactNode;
  rightIcon?: React.ReactNode;
  isDisabled?: boolean;
  hasError?: boolean;
  customClassName?: string;
  onChange: React.ReactEventHandler<HTMLInputElement>;
  localizationOptions?: InputLocalizationOptions;
};
```

rollup.config.js
```diff
@@ -31,8 +31,10 @@ export default [
       Textarea: "src/form/textarea/Textarea.tsx",
       Toggle: "src/toggle/Toggle.tsx",
       Switch: "src/switch/Switch.tsx",
-      DateTimer: "src/date-timer/DateTimer.tsx",
-      ProgressBar: "src/progress-bar/ProgressBar.tsx"
+      Countdown: "src/countdown/Countdown.tsx",
```
mucahit, line 34: Let's remove this, we renamed it as `DateTimer`

stories/18-Time.stories.tsx
```diff
@@ -0,0 +1,69 @@
+import {storiesOf} from "@storybook/react";
```
mucahit, line 1: Let's rename this file as `18-TimeInput.stories.tsx`

mucahit (changes requested):

src/core/utils/time/timeUtils.ts
```diff
@@ -42,4 +50,182 @@ function calculateRemainingTimeBreakdown(
...
+  } = options || {};
+
+  return (date: Date): string => {
+    let dateToFormat = date;
+
+    if (timeZone && isProvidedDateInUTC) {
+      dateToFormat = utcToZonedTime(date, timeZone);
+    } else if (!timeZone && shouldShiftDateToCompensateForTimezone) {
+      dateToFormat = compansateForTimezone(date);
+    }
```
akgunberk, line 85:
```suggestion
if (timeZone) {
    if (isProvidedDateInUTC) dateToFormat = utcToZonedTime(date, timeZone);
  } else if (shouldShiftDateToCompensateForTimezone) {
    dateToFormat = compansateForTimezone(date);
  }
```
as a kind suggestion:
I am not sure exactly of my suggestion but this look more readable for me because isProvidedDateInUTC is only matters if timezone provided I guess 🤔 I might miss some point ofc 😇

akgunberk (commented):
looks great overall! thanks 🥳

Anlerkan (approved):

mucahit (approved):

akgunberk (approved):

package.json
```diff
@@ -102,6 +102,8 @@
   },
   "dependencies": {
     "classnames": "2.3.1",
+    "date-fns": "2.21.1",
+    "date-fns-tz": "1.1.4",
```
edizcelik, line 106: There are newer versions available on npm for these now. Let's upgrade to them.

edizcelik (approved):

edizcelik (approved):

Issue #62, opened 2021-04-06
Button: Width should be dynamically determined depending on the text


edizcelik:
We currently predefine the width of the buttons (i.e. setting the width as 120px with CSS), however, we should instead set padding of buttons and let the element decide what its width would be depending on the provided children to Button component and the defined paddings. When displaying the spinner, the button should stay at the same width as if the text was visible instead of shrinking its width down.

PR #84, opened 2021-04-22
feat(button): Adjust dynamically button width

yasincaliskan:
### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/62

image


src/button/Button.tsx
```diff
@@ -73,16 +84,9 @@ function Button({
...
-        <Spinner
-          customClassName={classNames("button__spinner", {
-            [`${customClassName}__spinner`]: customClassName
-          })}
-          aria-label={"Button spinner visible. Button inactivated."}
-        />
-      ) : (
-        children
-      )}
+      <div className={childrenClassName}>{children}</div>
```
edizcelik, line 87: Adding a wrapper div to the children might require many changes in the projects that use Button component. I wonder if there is a way to hide the children of `button` tag that is not the spinner, `not` utility of CSS may help:

```
.button--is-pending {
   & > *:not(.button__spinner)
}
```

stories/3-Button.stories.tsx
```diff
@@ -26,6 +27,19 @@ storiesOf("Button", module).add("Button States", () => (
...
+    <StateProvider initialState={false}>
+      {(state, setState) => (
+        <Button
+          type={"button"}
+          onClick={() => setState(true)}
+          shouldDisplaySpinner={state}>
+          {"Click Me - shouldDisplaySpinner"}
+        </Button>
+      )}
+    </StateProvider>
```
edizcelik, line 41: Do we need this one? We already have an example for `shouldDisplaySpinner` above

src/button/Button.tsx
yasincaliskan: You are right. But when we add this CSS, `.button__spinner` isn't working correctly. As we know, I guess it has not specific enough. However, it's a little bit hacky but when we use like following, looks working. 
```
.button--is-pending {
  color: var(--button-bg);
}
```

src/button/_button.scss
```diff
@@ -31,8 +31,20 @@
   cursor: not-allowed;
 }
 
+.button--is-pending {
+  color: var(--button-bg);
+}
+
 .button__spinner {
+  position: absolute;
+
```
jamcry, line 40: I think we should also set button as relative:
```scss
.button {
  position: relative;
...
}
```

src/button/_button.scss
```diff
@@ -31,8 +31,20 @@
   cursor: not-allowed;
 }
 
+.button--is-pending {
+  color: var(--button-bg);
+}
```
jamcry, line 36: I have read the conversation above, I think this may cause strange bugs. How about setting `opacity: 0`? Then we can make sure it is not visible.

Also, you said spinner wasn't working for the Ediz's method. Maybe it can be because `.button` is not relatively positioned (see my other comment)

src/button/_button.scss
```diff
@@ -31,8 +31,20 @@
...
   .spinner__background {
     background: var(--button-bg);
   }
 }
+
+.button__children {
+  &--is-hidden {
+    visibility: hidden;
+  }
+}
```
jamcry, line 50: I guess we can remove this because the wrapper div is removed
```suggestion
```

src/button/Button.tsx
edizcelik: This may lead to strange bugs, we should find a better solution.

src/button/_button.scss
yasincaliskan: I've tried again but unfortunately, the result is the same. Also, when we set `opacitiy: 0`, the wrapper is disappearing with the background and all of them.

I guess we need to declare to children, not button classes for `opacity` or `visibility` properties.

src/dropdown/_dropdown.scss
```diff
@@ -58,6 +58,7 @@ $dropdown-header-horizontal-padding: 8px;
   display: flex;
   align-items: center;
   justify-content: space-between;
+  gap: 8px;
```
mucahit, line 61: This is not working on Safari (For flexbox), let's use `margin` instead

bilgeyucel (approved):
Other than the ongoing discussion, looks nice 👍

src/button/Button.tsx
```diff
@@ -76,10 +75,12 @@ function Button({
       {children}
 
       {shouldDisplaySpinner && (
-        <Spinner
-          customClassName={"button__spinner"}
-          aria-label={"Button spinner visible. Button inactivated."}
-        />
+        <div className={"button__spinner__container"}>
```
jamcry, line 78: I think the class name should be `button__spinner-container` (meaning `spinner-container` inside `button`)
If we use `__`, it looks like container is the child of spinner,
and '--' is for modifiers (`--is-active` etc.),
so `-` looks more suitable here

yasincaliskan: Yes, you're right. As Ediz said, we can use `button__spinner-container` then.

jamcry (approved):

edizcelik (approved):

Anlerkan (approved):

src/button/_button.scss
```diff
@@ -31,6 +33,14 @@
   cursor: not-allowed;
 }
 
+.button__spinner-container {
+  position: absolute;
```
mucahit, line 37: I'm not sure but probably we need to set `top,left,right,bottom to 0` make sure it's on correct position

mucahit (approved):

PR #85, opened 2021-04-23
feat(input): Add number input localization utility

yasincaliskan:
### Decription
- `numberToString` utility has added. Locale parameter is `"en"` as default.
- If `shouldFormatToLocaleString` is true, `numberToString` function trigger and return the converted value.
- Integer part corverts with before dot operator.

image


edizcelik (changes requested):
Didn't go through the PR in detail but I have a few questions:

- Can't I specify a specific locale via a prop? For example, can I force all number inputs to work in the “TR” locale?
- I think we should use `Intl.NumberFormat` rather than a custom `numberToString` utility? Otherwise, this wouldn't work properly for all locales. Let's try to implement this using `Intl.NumberFormat`
- I wonder if we should remove  `shouldFormatToLocaleString` prop altogether and default to using system locale if the `locale` prop was not provided to the Input.
- `handleChange` seems to assume the thousandths and decimal separator to be specific strings, `,` and `.` respectively. This is not true for all locales. For example, in "tr" locale, the number formatting looks something like "25.010,75" instead of "25,010.75" like in "en" locale.

Please try this on your local by changing your system locale to different countries that uses different number formatting. In addition, you should test it by providing explicit locales that uses different number formatting as well.

edizcelik:
This may help: https://github.com/globalizejs/globalize/blob/master/doc/api/number/number-parser.md but I think we will need to strict the locales.

edizcelik:
This [answer](https://stackoverflow.com/a/55366435) seems promising

yasincaliskan:
Thank you for your suggestions!
- I think we shouldn't remove `shouldFormatToLocaleString` prop. Because we can use `NumberInput` for different purposes like phone number or ID. And we can add a `locale` prop but if the user doesn't pass this prop, we can use the system locale as default. I don't think it will cause any problem.
- `Intl.NumberFormat` is working fine but `format` method assumes that taking like `123456.05` number. `formatToParts` method is very useful but I couldn't combine this method with our RegExp yet. I guess there is some weird stuff 🤔 
- I believe that, if we have the proper `DECIMAL_NUMBER_SEPARATOR` and `THOUSANDTHS_SEPARATOR` we can handle with `numberToString` function. Because I think so only `Intl.NumberFormat` not enough.

src/core/utils/number/numberUtils.ts
```diff
@@ -0,0 +1,86 @@
...
+    // If the value is `"[0-9]."` or `[0-9].0+"
+    else if (
+      IS_LAST_CHARACTER_DECIMAL_POINT_REGEX.test(value) ||
+      MATCH_ZEROS_AFTER_DECIMAL_REGEX.test(value) ||
+      value.includes(DECIMAL_NUMBER_SEPARATOR)
+    ) {
+      const decimalNumberParts = value.split(DECIMAL_NUMBER_SEPARATOR);
+      const decimalPart = decimalNumberParts[1];
+      const integerPart = parseFloat(decimalNumberParts[0]).toLocaleString(locale);
+
```
edizcelik, line 38: I thought we would be using Intl.NumberFormat directly instead of doing the calculations below. Rename the function to `formatNumber` and the implementation may look like this:

```
function formatNumber(
  providedOptions: Omit<Intl.NumberFormatOptions, "style"> & {
    style?: Intl.NumberFormatPartTypes | "percent";
    locale?: string;
  } = {}
) {
  const {locale, ...otherOptions} = providedOptions;
  let options = otherOptions;
  const isCurrencyFormatting = options.style === "currency";

  if (isCurrencyFormatting) {
    options = {
      currency: "USD",
      maximumFractionDigits: 5,
      minimumFractionDigits: 2,
      // "narrowSymbol" prevents showing explicit currency symbols, such as "US$" or "CA$"
      currencyDisplay: "narrowSymbol",
      ...otherOptions
    };
  } else {
    options = {
      maximumFractionDigits: 5,
      style: "decimal",
      ...otherOptions
    };
  }

  let numberFormatter: {
    format: (x: number | bigint) => string;
  };

  try {
    numberFormatter = new Intl.NumberFormat(
      locale || [navigator.language, "en-GB"],
      options
    );
  } catch (error) {
    numberFormatter = {
      format(x: number | bigint) {
        return x.toLocaleString(locale);
      }
    };

    if (isCurrencyFormatting) {
      // currencyDisplay: "narrowSymbol" option is not supported by some browsers, try by providing "symbol"
      try {
        options.currencyDisplay = "symbol";
        numberFormatter = new Intl.NumberFormat(
          locale || [navigator.language, "en-GB"],
          options
        );
      } catch (currencyFormattingError) {
        console.error({currencyFormattingError});
      }
    }
  }

  return (value: number) => {
    let formattedValue = "";

    if (!Object.is(value, NaN)) {
      formattedValue = numberFormatter.format(value);
    }

    if (formattedValue && options.currency === "USD") {
      // in case `narrowSymbol` option was failed, make sure "US$" sign appears as just "$"
      formattedValue = formattedValue.replace("US$", "$");
    }

    return formattedValue;
  };
}
```

src/core/utils/number/numberUtils.ts
```diff
@@ -0,0 +1,86 @@
...
+          .replace(DECIMAL_NUMBER_SEPARATOR, "")
+          .replace(THOUSANDTHS_SEPARATOR, "")}`;
+      }
+    }
+  }
+
+  return final;
+}
+
+function parseNumber(value: number | string, locale = navigator.language) {
```
edizcelik, line 57: Let's add documentation for this function as well. Also, please provide a link to the Observable post we used as a reference here.

src/core/utils/number/numberUtils.ts
```diff
@@ -0,0 +1,86 @@
...
+    .replace(numeral, digitMapper);
+
+  function getDigit(digitMap: Map<string, number>) {
+    return (d: string) => {
+      const digit = digitMap.get(d);
+
+      return typeof digit === "number" ? String(digit) : "";
+    };
+  }
+}
```
edizcelik, line 85:
```suggestion
}

  function getDigit(digitMap: Map<string, number>) {
    return (d: string) => {
      const digit = digitMap.get(d);

      return typeof digit === "number" ? String(digit) : "";
    };
  }
```

src/form/input/Input.tsx
```diff
@@ -118,7 +130,11 @@ function Input(props: InputProps) {
       if (maxFractionDigits > 0) {
         const decimalNumberParts = newValue.split(DECIMAL_NUMBER_SEPARATOR);
         const decimalPart = decimalNumberParts[1];
-        const integerPart = decimalNumberParts[0].replace(THOUSANDTHS_SEPARATOR, "");
+        const integerPart = decimalNumberParts[0];
```
edizcelik, line 133: As it was discussed during the weekly meeting, `onChange` handler should `parse` the input value into a real number. So I believe `formattedNewValue` should just be computed using `parseNumber` utility

src/form/input/Input.tsx
```diff
@@ -75,6 +81,11 @@ function Input(props: InputProps) {
     "input--has-error": hasError
   });
   const isNumberInput = type === "number";
+  let finalValue = value;
+
+  if (isNumberInput && value && shouldFormatToLocaleString) {
+    finalValue = numberToString(parseNumber(String(value)), maxFractionDigits, locale);
```
edizcelik, line 87: Here we will FORMAT the number into the localized string representation with "formatNumber" function

stories/0-Form.stories.tsx
```diff
@@ -206,6 +209,81 @@ storiesOf("Form", module)
...
+                value={state}
+              />
+            </FormField>
+
+            <p>{`TR Locale: ${numberToString(state, 4, "tr-TR")}`}</p>
+            <p>{`CN Locale: ${numberToString(state, 4, "zh-Hans-CN-u-nu-hanidec")}`}</p>
+            <p>{`AR Locale: ${numberToString(state, 4, "ar-EG")}`}</p>
+          </Fragment>
+        )}
+      </StateProvider>
```
edizcelik, line 286: Let's provide an example with `locale` prop

stories/0-Form.stories.tsx
```diff
@@ -206,6 +209,81 @@ storiesOf("Form", module)
...
+              <Input
+                name={"world-population"}
+                type={"number"}
+                placeholder={"7,794,798,739"}
+                shouldFormatToLocaleString={true}
+                onChange={(e) => setState(e.currentTarget.value)}
+                value={state}
+              />
+            </FormField>
+          </Fragment>
```
edizcelik, line 256:
```suggestion
          <p>{`event.currentTarget.value: ${state}`}</p>
          </Fragment>
```

stories/0-Form.stories.tsx
```diff
@@ -206,6 +209,81 @@ storiesOf("Form", module)
...
+                name={"wallet"}
+                maxFractionDigits={2}
+                type={"number"}
+                placeholder={"$ 1,000"}
+                shouldFormatToLocaleString={true}
+                onChange={(e) => setState(e.currentTarget.value)}
+                value={state}
+              />
+            </FormField>
+          </Fragment>
```
edizcelik, line 234:
```suggestion
          <p>{`event.currentTarget.value: ${state}`}</p>
          </Fragment>
```

edizcelik (changes requested):

src/core/utils/number/numberUtils.ts
yasincaliskan: Thank you for your help! 🙏🏻

src/form/input/Input.tsx
```diff
@@ -49,32 +51,61 @@ export type InputProps = Omit<
...
+
+    if (
+      IS_LAST_CHARACTER_DECIMAL_POINT_REGEX.test(String(value)) ||
+      MATCH_ZEROS_AFTER_DECIMAL_REGEX.test(String(value))
+    ) {
+      const decimalPart = String(value).split(DECIMAL_NUMBER_SEPARATOR)[1] || "";
+      const integerPart = finalValue.split(DECIMAL_NUMBER_SEPARATOR)[0];
+
+      finalValue = `${integerPart}${DECIMAL_NUMBER_SEPARATOR}${decimalPart}`;
+    }
```
edizcelik, line 107: whats this for ?

src/form/input/Input.tsx
```diff
@@ -109,21 +141,24 @@ function Input(props: InputProps) {
...
+        IS_LAST_CHARACTER_DECIMAL_POINT_REGEX.test(formattedNewValue) ||
+        MATCH_ZEROS_AFTER_DECIMAL_REGEX.test(formattedNewValue)
+      ) {
+        const decimalNumberParts = formattedNewValue.split(DECIMAL_NUMBER_SEPARATOR);
         const decimalPart = decimalNumberParts[1];
-        const integerPart = decimalNumberParts[0].replace(THOUSANDTHS_SEPARATOR, "");
+        const integerPart = decimalNumberParts[0];
+
+        formattedNewValue = `${integerPart}${DECIMAL_NUMBER_SEPARATOR}${decimalPart}`;
+      }
```
edizcelik, line 155: I don't think we require this part any longer

src/form/input/Input.tsx
```diff
@@ -109,21 +141,24 @@ function Input(props: InputProps) {
...
-        if (decimalPart && decimalPart.length > maxFractionDigits) {
-          const trimmedDecimalPart = decimalPart.slice(0, maxFractionDigits);
+      if (maximumFractionDigits > 0) {
+        const decimalPart = newValue.split(DECIMAL_NUMBER_SEPARATOR)[1];
 
-          formattedNewValue = `${integerPart}${DECIMAL_NUMBER_SEPARATOR}${trimmedDecimalPart}`;
+        if (decimalPart && decimalPart.length === maximumFractionDigits + 1) {
+          return;
         }
       } else {
```
edizcelik, line 163: I think we can provide `maximumFractionDigits` to `parseNumber` as an option and handle this there.

src/core/utils/number/numberUtils.ts
```diff
@@ -0,0 +1,80 @@
...
+  };
+}
+
+/**
+ * Coerces a number scientific notation. {@link https://observablehq.com/@mbostock/localized-number-parsing | Reference}
+ * @param {number} value - A number to convert to string
+ * @param {string} locale - Default locale used is browser locale
+ * @returns {string} The value after coercing the given value to a scientific notation.
+ */
+function parseNumber(value: number | string, locale = navigator.language) {
```
edizcelik, line 46: Let's swap the order of the parameters and turn the first parameter into an "options" object
```suggestion
function parseNumber(options: ParseNumberOptions = {locale: navigator.language}, value: number | string) {
```

src/form/input/Input.tsx
```diff
@@ -109,21 +141,24 @@ function Input(props: InputProps) {
...
         formattedNewValue = formattedNewValue.replace(
```
edizcelik, line 164: We can move this to `parseNumber` as well and make sure that it does not to return "1000." kinda values. Then, we can remove this else block too. So, we should move all the number parsing related code to `parseNumber` and this handler should look something like this in the end:

```
    if (isNumberInput) {
      const {value: newValue} = event.currentTarget;
      const formattedNewValue = parseNumber(newValue, locale);

      if (formattedNewValue !== newValue) {
        event.currentTarget.value = formattedNewValue;
      }
    }
```

edizcelik (changes requested):

src/form/input/Input.tsx
yasincaliskan: `formatNumber` expects a number for value and we convert it to number. (line: 97)

Meanwhile `decimal separator` and zeros disappear. I think we should hold these. 
```js
const stringNum = "100.0";
const convertedNum = Number(stringNum); // 100 as number
```

edizcelik: ha okay makes sense

src/form/input/Input.tsx
yasincaliskan: When we remove this condition, the user can enter any character. I guess we need this condition both for formatting and parsing. Maybe we can create a new utility for this. What do you think?

edizcelik: Hmm. Not sure I understand. Doesn't this just check if the entered value is a fraction and then re-construct the value in natural number format ? What characters are we trying to prevent users to use with this if block ?

yasincaliskan: I think users can enter only number and decimal operator.

edizcelik: How do we check against the decimal operator ?

src/form/input/Input.tsx
edizcelik: Unresolving this because I have another question:

So let's say the `value` is scientific notation "12345.0" at the moment because user eventually wants to write "12345.005" into the input. To call `formatNumber`, we run `Number("12345.0")` and then `formatNumber` returns us `12.345` (in the TR locale). However, we want the input to read `12.345,0` because the value is supposed to be `"12345.0"`. To achieve this, we run this if block:

```
    if (
      IS_LAST_CHARACTER_DECIMAL_POINT_REGEX.test(String(value)) ||
      MATCH_ZEROS_AFTER_DECIMAL_REGEX.test(String(value))
    ) {
      const decimalPart = String(value).split(DECIMAL_NUMBER_SEPARATOR)[1] || "";
      const integerPart = finalValue.split(DECIMAL_NUMBER_SEPARATOR)[0];

      finalValue = `${integerPart}${DECIMAL_NUMBER_SEPARATOR}${decimalPart}`;
    }
 ```
 
 now the finalValue becomes `12345,0` in the TR locale. However, it should be `12.345,0`, should it not? Did I miss something ?

File: src/form/input/Input.tsx (hunk 10 above)
Comment (edizcelik) [reply src/form/input/Input.tsx:155]: Nevermind. We are testing against `formattedNewValue` which is in scientific format so `IS_LAST_CHARACTER_DECIMAL_POINT_REGEX` and `MATCH_ZEROS_AFTER_DECIMAL_REGEX` works properly here.

Comment (edizcelik) [reply src/form/input/Input.tsx:155]: > When we remove this condition, the user can enter any character

What does `parseNumber` return if the event.currentTarget.value is something like "10a" ?

Comment (yasincaliskan) [reply src/form/input/Input.tsx:155]: `event.currentTarget.value` is returning `"10a"` in this version. I think we should prevent entering other characters without number or decimal separator in the handler.

I will think how we can combine your above suggestion with this.

File: src/form/input/Input.tsx (hunk 9 above)
Comment (yasincaliskan) [reply src/form/input/Input.tsx]: Great, thank you!

**Comment by yasincaliskan** [comment]
Looks like working well with your help. Thank you 🎉

But there is one more thing. We should think of the following `replace` scope for other locales. It's working fine with only`0123456789`.
```js
let parsedNumber = value
.replace(new RegExp(NOT_NUMBER_OR_DECIMAL_POINT_REGEX), "")
...
```

It couldn't work for any different number of characters. (see last story example)
Have you an idea for this?

src/core/utils/number/numberUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+    .replace(numeral, digitMapper);
+
+  if (options.maximumFractionDigits > 0) {
+    const decimalPart = parsedNumber.split(DECIMAL_NUMBER_SEPARATOR)[1];
+
+    if (decimalPart && decimalPart.length === options.maximumFractionDigits + 1) {
+      return parsedNumber.slice(0, parsedNumber.length - 1);
+    }
+  } else {
+    parsedNumber = parsedNumber.replace(IS_LAST_CHARACTER_DECIMAL_POINT_REGEX, "");
```
edizcelik, line 76: I think here we want to make sure the parsed number is an integer so I wonder if we can do this instead:
```suggestion
    parsedNumber = String(parseInt(parsedNumber));
```

src/core/utils/number/numberUtils.ts
```diff
@@ -0,0 +1,99 @@
...
+    .replace(new RegExp(NOT_NUMBER_OR_DECIMAL_POINT_REGEX), "")
+    .replace(new RegExp(`[${THOUSANDTHS_SEPARATOR}]`, "g"), "")
+    .replace(new RegExp(`[${DECIMAL_NUMBER_SEPARATOR}]`), ".")
+    .replace(numeral, digitMapper);
+
+  if (options.maximumFractionDigits > 0) {
+    const decimalPart = parsedNumber.split(DECIMAL_NUMBER_SEPARATOR)[1];
+
+    if (decimalPart && decimalPart.length === options.maximumFractionDigits + 1) {
+      return parsedNumber.slice(0, parsedNumber.length - 1);
```
edizcelik, line 73: why are we returning an array in this case ?

src/form/input/Input.tsx
```diff
@@ -49,32 +51,67 @@ export type InputProps = Omit<
...
+  let finalValue = value;
+
+  if (isNumberInput && value && shouldFormatToLocaleString) {
+    const numberFormatter = formatNumber({
+      providedOptions: {
+        maximumFractionDigits,
+        locale
+      }
+    });
+
```
edizcelik, line 100: Can you add a comment here and describe why we require these lines?

src/core/utils/number/numberUtils.ts
yasincaliskan: Actually, `slice` returns a string. Is it a wrong approach? Also, this check not enough for like `1234.10.` format. We don't prevent the second decimal separator now. I'll think about it.

edizcelik: 🤦‍♂️ you are right, I don't know why I thought we were returning an array here. For additional decimal separators, I think we should prevent them

src/form/input/Input.tsx
```diff
@@ -49,32 +48,77 @@ export type InputProps = Omit<
...
   const inputClassName = classNames("input", {
     "input--is-disabled": isDisabled,
     "input--has-error": hasError
   });
   const isNumberInput = type === "number";
+  let finalValue = value;
+
+  if (isNaN(Number(value))) {
+    finalValue = "";
+  }
```
yasincaliskan, line 93: In the current version, `parseNumber` returns `NaN` that are not numbers. I've tried `event.preventDefault()` in `handleChange` but it's not working. (cc: @edizcelik) You can take a look [here](https://stackoverflow.com/questions/32298064/preventdefault-not-working-in-on-input-function). 
Therefore, I thought we can handle like that but I'm not sure it's good solution.

edizcelik: We need to handle this within `handleChange`. If `formattedNewValue` is NaN after `parseNumber` is called, then we can preventDefault. we should be able to use preventDefault for onChange handlers, not sure why it isn't working. If it is really not working, I assume we can revert back to the latest `value` (or the computed finalValue) that is provided as a prop .

yasincaliskan: I've investigated but it seems there is no way use `event.preventDefault()` for input event. I've tried for `SyntheticEvent`, `KeyboardEvent`, `ChangeEvent` but `HTMLInputElement` have not `key` property for use `preventDefault`. 
I guess as you said, we should use previous state.

edizcelik: We can just not call the `onChange` handler in this case so that the value wouldn't be updated. Would that work ?

yasincaliskan: I've tried a similar one but didn't work. I guess really correct a RegExp will solve this. Even it's working now but in dynamic format causes some weird behaviors.

edizcelik: Does the check on line:91 work ? If so, you can just do the same within `handleChange`

yasincaliskan: It's working only in one case that hasn't `maximumFractionDigits` prop.
I believe that if we can work `^-?([0-9]*,*)*\\.?[0-9]*$` Regex as dynamically, it will work. But somehow it's not working. I guess I make a mistake in some operators.

Shouldn't be like this?
```js
const NUMERALS = "[0123456789]"; // dynamically
const NUMBER_WITH_THOUTHOUSANDTH_AND_DECIMAL_POINT_REGEX = new RegExp(
 `^-?(${NUMERALS}|[0-9]*${THOUSANDTHS_SEPARATOR}*)*\\${DECIMAL_SEPARATOR}?${NUMERALS}|[0-9]*$`
);
```

src/form/input/Input.tsx
```diff
@@ -49,32 +47,77 @@ export type InputProps = Omit<
...
     ...rest
   } = props;
+  const {
+    shouldFormatToLocaleString = false,
+    locale,
+    maximumFractionDigits = 0
+  } = localizationOptions;
+  const [decimalSeparatorForLocale, setDecimalSeparatorForLocale] = useState(
+    () => getNumberSeparators(locale).DECIMAL_NUMBER_SEPARATOR
+  );
```
edizcelik, line 83: As we are using both `DECIMAL_NUMBER_SEPARATOR` and `THOUSANDTHS_SEPARATOR`, let's store the whole thing within the state and then them below

```suggestion
  const [{
    DECIMAL_NUMBER_SEPARATOR: decimalSeparatorForLocale,
    THOUSANDTHS_SEPARATOR: thousandthSeparatorForLocale
  }, setNumberSeparatorsForLocale] = useState(
    () => getNumberSeparators(locale)
  );
```

edizcelik (changes requested):

edizcelik (changes requested):

bilgeyucel (approved):
🚀

src/core/utils/number/numberUtils.ts
```diff
@@ -0,0 +1,100 @@
...
+  try {
+    numberFormatter = new Intl.NumberFormat(
+      locale || [navigator.language, "en-GB"],
+      options
+    );
+  } catch (error) {
+    numberFormatter = {
+      format(x: number | bigint) {
+        return x.toLocaleString(locale);
+      }
```
jamcry, line 23: Can we add a comment here to describe what kind of error we are expecting?
I think this `catch` block handles the cases where `NumberFormat` (or `navigator`) is not supported by the browser, if so, we can add something like:
`// NumberFormat or navigator is not supported, just use toLocaleString for formatting`

src/core/utils/number/numberUtils.ts
```diff
@@ -0,0 +1,100 @@
...
+  const DECIMAL_NUMBER_SEPARATOR = parts.find((d) => d.type === "decimal")!.value;
+
+  return {THOUSANDTHS_SEPARATOR, DECIMAL_NUMBER_SEPARATOR};
+}
+
+function getLocaleNumerals(locale = navigator.language) {
+  const numerals = [
+    // eslint-disable-next-line no-magic-numbers
+    ...new Intl.NumberFormat(locale, {useGrouping: false}).format(9876543210)
+  ].reverse();
```
jamcry, line 95:
```suggestion
    // eslint-disable-next-line no-magic-numbers
  const numerals = new Intl.NumberFormat(locale, {useGrouping: false}).format(9876543210).split("").reverse();
```
I think this version may be easier to read. I had a hard time making sense of this block (I didn't know we can split string by destructring 😄 )

jamcry (approved):
Looks good overall, thanks 👏  Left a few suggestions.

edizcelik (approved):

Anlerkan (approved):

src/core/utils/number/numberUtils.ts
yasincaliskan: Yes, more descriptive. Thank you!

src/core/utils/number/numberUtils.ts
yasincaliskan: Looks better!

PR #87, opened 2021-04-27
fix(root): Fix eslint issues

yasincaliskan:
### Description
- Some eslint rule violations were ignored.
- `"react/jsx-handler-names"` rule was added to `.eslintrc`.
- `.eslintrc` was added to `.eslintignore`.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -28,7 +28,8 @@ export default [
       FileUploadButton: "src/button/file-upload/FileUploadButton.tsx",
       Spinner: "src/spinner/Spinner.tsx",
       Countdown: "src/countdown/Countdown.tsx",
-      ProgressBar: "src/progress-bar/ProgressBar.tsx"
+      ProgressBar: "src/progress-bar/ProgressBar.tsx",
+      Textarea: "src/form/textarea/Textarea.tsx"
     },
     output: {
       dir: "dist",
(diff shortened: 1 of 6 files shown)


rollup.config.js
```diff
@@ -28,8 +28,7 @@ export default [
       FileUploadButton: "src/button/file-upload/FileUploadButton.tsx",
       Spinner: "src/spinner/Spinner.tsx",
       Countdown: "src/countdown/Countdown.tsx",
-      ProgressBar: "src/progress-bar/ProgressBar.tsx",
-      Textarea: "src/form/textarea/Textarea.tsx"
```
jamcry, line 32: Was this a mistake? Why are we removing `Textarea` ?

bilgeyucel (approved):
Looks good! Apart from these updates, let's open an issue to update eslint package versions

rollup.config.js
yasincaliskan: Ops! I supposed that there is duplicated `Textarea` but there wasn't. My bad, sorry! 😅 

Actually, we can provide finding eslint issues while npm is running. Should we add a script to `package.json` for this?

jamcry: Yes, I think we should add a step to build command for linting. This way, we can make sure there are no errors before publishing to npm.
I think we can create an issue (or update, if you opened one for the issue Bilge mentioned below), and add it later.

jamcry (approved):

edizcelik (approved):

mucahit (approved):

PR #91, opened 2021-05-07
fix(core): Add `--isolatedModules` flag

yasincaliskan:
- `--isolatedModules` flag was added.
- Prop export type was changed.

```
Re-exporting a type when the '--isolatedModules' flag is provided requires using 'export type'.
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,29 +1,58 @@
 import "./ui/reference/_colors.scss";
 import "./ui/reference/_measurement.scss";
 
-import FormField from "./form/field/FormField";
-import Input from "./form/input/Input";
-import PasswordInput from "./form/password-input/PasswordInput";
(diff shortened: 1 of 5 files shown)


src/index.ts
```diff
@@ -57,33 +49,34 @@ export {
...
-  TextareaProps,
-  ToggleProps,
-  SwitchProps,
-  CountdownProps,
   // Hooks
   useCountDownTimer
 };
+
+// Types
+export type FormFieldProps = any;
```
edizcelik, line 57:
```suggestion
export type FormFieldProps = FormFieldProps;
```

We should be exporting the types otherwise we wont be able to read them within projects

Anlerkan (approved):

mucahit (approved):

PR #95, opened 2021-05-12
feat(typeahead): Add type prop for TypeaheadInput

yasincaliskan:
- `typeaheadProps` can use all input types.

image

### DIFF
--- a/src/form/input/typeahead/TypeaheadInput.tsx
+++ b/src/form/input/typeahead/TypeaheadInput.tsx
@@ -1,19 +1,20 @@
 import React, {useEffect} from "react";
 import classNames from "classnames";
 
-import Input from "../Input";
+import Input, {InputTypes} from "../Input";
 import useDebounce from "../../../core/utils/hooks/debounce";
 
 export interface TypeaheadInputProps {
   onQueryChange: (value: string) => void;
+  name: string;
+  placeholder: string;
+  type?: InputTypes;
   value?: string;
   testid?: string;
   customClassName?: string;
   id?: string;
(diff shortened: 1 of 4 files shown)


edizcelik (approved):
Cool. Thanks @yasincaliskan

jamcry (approved):

Anlerkan (approved):

mucahit (approved):

src/form/input/typeahead/TypeaheadInput.tsx
```diff
@@ -1,19 +1,20 @@
...
 
-import Input from "../Input";
+import Input, {InputTypes} from "../Input";
 import useDebounce from "../../../core/utils/hooks/debounce";
 
 export interface TypeaheadInputProps {
   onQueryChange: (value: string) => void;
+  name: string;
+  placeholder: string;
+  type?: InputTypes;
```
bilgeyucel, line 11: I think the `type` option should only accept searchable types such as`text` and `number`, not all `InputTypes`. It would look really weird if I give `color` as type and there is no real-life scenario to use it.

yasincaliskan: You're right, color type doesn't make sense. We use text type basically. We can remove except `text` and `number` types.

edizcelik: Ok. @yasincaliskan let's use `Extract` util on InputTypes to generate a union type with only "text" and "number".

edizcelik (approved):

PR #96, opened 2021-05-19
feat: Initialize tests

Anlerkan:
### Firstly Merge

- #103 , #104 , #113 , #136, #143

### Description
- Jest configuration improved
- Install  `jest-axe` and `axe-core` for `a11y` tests

### Current Status
- X: ---> Hasn't started yet
- ✓: ---> Completed -Ready for Review-
- ✓~: ---> W.I.P  
```
└── src
    └── avatar X
    └── button ✓
    └── core 
        └── utils X
    └── countdown X
    └── dropdown X
    └── form
        └── field ✓
        └── input ~✓
            └── checkbox ✓
            └── file ✓
            └── radio ✓
            └── typeahead ✓~
        └── password-input ✓
        └── textarea ✓
    └── list ✓ #103 
    └── progress-bar ✓ #104 
    └── select X
    └── spinner X
    └── switch X
    └── tab X
    └── tag X
    └── toggle X
```


yasincaliskan:
Thank you, seems awesome! 🎉

src/button/button.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+        shouldDisplaySpinner={false}
+        customSpinner={customSpinnerContent}
+        {...defaultButtonProps}
+      />
+    );
+
+    expect(getByTestId("button")).not.toContainElement(customSpinner);
+  });
+
+  it("should not call click event when button is disabled", () => {
```
edizcelik, line 74:
```suggestion
  it("should not run click event handler while button is disabled", () => {

```

src/button/button.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+      <Button onClick={handleClick} shouldDisplaySpinner={true} {...defaultButtonProps} />
+    );
+
+    fireEvent.click(getByTestId("button"));
+    expect(handleClick).not.toHaveBeenCalled();
+
+    rerender(<Button onClick={handleClick} isDisabled={true} {...defaultButtonProps} />);
+
+    fireEvent.click(getByTestId("button"));
+    expect(handleClick).not.toHaveBeenCalled();
```
edizcelik, line 87: I think we can separate these two as one is for not being able to run the handler when `shouldDisplaySpinner` is true while the other is for testing the case where `isDisabled` is true

src/button/button.test.tsx
```diff
@@ -0,0 +1,154 @@
...
+  });
+
+  it("should call events correctly", () => {
```
edizcelik, line 90: Wouldn't it make more sense to test each event handler separately ?

Anlerkan:
As we spoke, I will open new PR for each new test . I'm marking this PR as `ready for review`   so that it wont get more complicated.

src/button/button.test.tsx
```diff
@@ -0,0 +1,178 @@
...
+    render(<Button {...defaultButtonProps} />);
+  });
+
+  it("should render children correctly", () => {
+    const {getByTestId} = render(<Button {...defaultButtonProps} />);
+
+    expect(getByTestId("button")).toHaveTextContent("Test");
+  });
+
+  it("should matches snapshot", () => {
```
edizcelik, line 28:
```suggestion
  it("should match snapshot", () => {

```

src/form/field/form-field.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+  const defaultFormFieldProps: FormFieldProps = {
+    testid: "form-field",
+    children: <Input onChange={jest.fn} name={"test"} testid={"form-field.input"} />
+  };
+
+  it("should render correctly", () => {
+    render(<FormField {...defaultFormFieldProps} />);
+  });
+
+  it("should matches snapshot", () => {
```
edizcelik, line 22:
```suggestion
  it("should match snapshot", () => {

```

src/form/field/form-field.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    await testA11y(container, {rules: {label: {enabled: false}}});
+  });
+
+  it("should render children correctly", () => {
+    const {getByTestId} = render(<FormField {...defaultFormFieldProps} />);
+
+    const childrenContent = getByTestId("form-field.input");
+
+    expect(getByTestId("form-field")).toContainElement(childrenContent);
```
edizcelik, line 39: Should we get the testid from `defaultFormFieldProps` ? If so, let's do it for other uses below as well. Also, should we add a test to see if the data-testid is added as an attribute when testid prop is provided or is that handled with the snapshot ?
```suggestion
    expect(getByTestId(defaultFormFieldProps.testid)).toContainElement(childrenContent);
```

src/form/field/form-field.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+    );
+
+    const formFieldLabel = document.getElementsByTagName("label")[0];
+
+    expect(getByTestId("form-field")).toHaveTextContent("Test");
+    expect(formFieldLabel).toHaveAttribute("id", "form-field.labelled-by");
+    expect(formFieldLabel).toHaveAttribute("for", "form-field.input");
+  });
+
+  it("errorMessages prop should work correctly", () => {
```
edizcelik, line 59:
```suggestion
  it("should handle errorMessages correctly", () => {

```

src/form/field/form-field.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    expect(getByTestId("form-field")).toHaveTextContent("Test Error 1");
+    expect(getByTestId("form-field")).toHaveTextContent("Test Error 2");
+
+    expect(getByTestId("form-field.error-messages").children).toHaveLength(
+      formFieldErrorMessages.length
+    );
+  });
+
+  it("helperMessages prop should work correctly", () => {
```
edizcelik, line 75:
```suggestion
  it("should handle helperMessages correctly", () => {

```

src/form/field/form-field.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+    const formFieldLabel = document.getElementsByTagName("label")[0];
+
+    expect(getByTestId("form-field")).toHaveTextContent("Test");
+    expect(formFieldLabel).toHaveAttribute("id", "form-field.labelled-by");
+    expect(formFieldLabel).toHaveAttribute("for", "form-field.input");
+  });
+
+  it("errorMessages prop should work correctly", () => {
+    const formFieldErrorMessages = ["Test Error 1", "Test Error 2"];
```
edizcelik, line 60: What's the behavior if I provide `[]` or `[""]` as errorMessages ? Same question for the helperMessages. Should we test these too ?

src/form/field/message/FormFieldMessage.tsx
```diff
@@ -1,27 +1,27 @@
...
-
-export default FormFieldMessage;
+import "./_form-field-message.scss";
+
+import React from "react";
+import classNames from "classnames";
+
+export interface FormFieldMessageProps {
+  className?: string;
+  type: "error" | "warning" | "helper";
```
edizcelik, line 8:
```suggestion
  type: "error" | "warning" | "helper";
  className?: string;
```

src/form/field/message/FormFieldMessage.tsx
```diff
@@ -1,27 +1,27 @@
...
-}
-
-export default FormFieldMessage;
+import "./_form-field-message.scss";
+
+import React from "react";
+import classNames from "classnames";
+
+export interface FormFieldMessageProps {
+  className?: string;
```
edizcelik, line 7: I wonder if we should rename this as `customClassName` just for consistency. What  do you think ?

src/form/field/message/FormFieldMessage.tsx
```diff
@@ -1,27 +1,27 @@
...
+
+function FormFieldMessage({className, type, message, testid}: FormFieldMessageProps) {
+  const messageRowClassName = classNames(
+    "form-field-message",
+    className,
+    `form-field-message--is-${type}`
+  );
+
+  return (
+    <p data-testid={testid} className={messageRowClassName}>
```
edizcelik, line 21: No need to define this variable
```suggestion
  return (
    <p data-testid={testid} className={classNames(
    "form-field-message",
    className,
    `form-field-message--is-${type}`
  )}>

```

src/form/field/message/form-field-message.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    await testA11y(container);
+  });
+
+  it("should render message correctly", () => {
+    const {getByTestId} = render(
+      <FormFieldMessage message={"Test Message"} {...defaultFormFieldMessageProps} />
+    );
+
+    expect(getByTestId("form-field-message")).toHaveTextContent("Test Message");
```
edizcelik, line 38:
```suggestion
    expect(getByTestId(defaultFormFieldMessageProps.testid)).toHaveTextContent("Test Message");
```

src/form/field/message/form-field-message.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+  });
+
+  it("should have proper class name", () => {
```
edizcelik, line 41:
```suggestion
  it("should have proper modifier class name for the type prop provided", () => {
```

src/form/field/message/form-field-message.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const defaultFormFieldMessageProps: FormFieldMessageProps = {
+    testid: "form-field-message",
+    type: "error"
+  };
+
+  it("should render correctly", () => {
+    render(<FormFieldMessage {...defaultFormFieldMessageProps} />);
+  });
+
+  it("should matches snapshot", () => {
```
edizcelik, line 21:
```suggestion
  it("should match snapshot", () => {
```

src/form/field/form-field.test.tsx
Anlerkan: @edizcelik  What is the expected behavior here? Right now, if i provide `empty array []` it does not render `error-message-list` but if i provide `[""]` it render an empty `FormFieldMessage` . Is this the expected behavior or do we need to make changes here?

src/form/field/form-field.test.tsx
Anlerkan: - Yes, we can get the testid from default props. 🙏 
-  I think we don't need to test `testid` because snapshot test already handle it.

src/form/input/file/file-unput.test.tsx
```diff
@@ -0,0 +1,113 @@
...
+    const fileInput = document.getElementsByTagName("input")[0]!;
+
+    fireEvent.change(fileInput, {target: {files}});
+
+    expect(fileInput.files).toHaveLength(files.length);
+    expect(fileInput.files![0]).toStrictEqual(files[0]);
+    expect(fileInput.files![1]).toStrictEqual(files[1]);
+    expect(defaultFileInputProps.onChange).toHaveBeenCalledTimes(files.length);
+  });
+
```
jamcry, line 71: As far as I can see, in these test cases, we are basically testing native `input` element. (Because we select input tag and fire event in this element) Since it is native browser element, testing seems redundant. I believe these can be omitted.

src/form/input/typeahead/typeahead-input.test.tsx
```diff
@@ -0,0 +1,32 @@
+import React from "react";
```
jamcry, line 1: I believe typeahead tests are not complete yet, right?

src/form/input/file/file-unput.test.tsx
Anlerkan: Makes sense. I removed this tests. Thank you

src/form/input/typeahead/typeahead-input.test.tsx
Anlerkan: Yes, sorry for the confusion

src/form/input/__snapshots__/input.test.tsx.snap
```diff
@@ -0,0 +1,17 @@
...
+>
+  <input
+    autoComplete="off"
+    autoCorrect="off"
+    className="input"
+    name="input"
+    onChange={[Function]}
+    type="text"
+  />
+</div>
```
yasincaliskan, line 16: Should we create snapshot instances for all input types? Our `<Input>` components is different than `<input>` element especially for `number` type.

mucahit (approved):

PR #97, opened 2021-05-20
fix(number-input): Decimal part improvements and number-input class

yasincaliskan:
- Add `number-input` class for `number` type
- Fix decimal part multiple zero issue

### Issue example:
`1,234.02` -> `1,234.2`


bilgeyucel (approved):

Anlerkan (approved):

edizcelik:
@yasincaliskan While I checked with the last example on Storybook I discovered that using parseInt or parseFloat on the "decimal part" is not producing the behavior we want. While trying it out some solutions, I thought the only consistent way of doing this would simply be mapping each digit manually to their localized version. I pushed a commit doing this.

Before:

Screen Shot 2021-05-21 at 12 46 10 PM

After using digit mapping:

Screen Shot 2021-05-21 at 1 32 06 PM

Now there seems to be another issue. When I test the examples with `maxFractionDigits={0}`, I can't seem to remove the value I entered to the input. For example, enter `1000` to the World Population and then try removing it. It doesn't remove `1`.

PR #98, opened 2021-05-21
fix(toast-provider): Allow control of autoClose via ToastProvider

edizcelik:
Some projects might want Toasts to not autoClose by default. This adds a prop to ToastProvider that allows this functionality. You can still pass `autoClose: true` with `toast.display` callback to override ToastProvider's autoCloseToasts prop value.

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -7,10 +7,5 @@ const req = require.context("../stories", true, /\.stories\.tsx$/);
 function loadStories() {
   req.keys().forEach(req);
 }
- 
-const ToastDecorator = storyFn => (
-  <ToastContextProvider>{storyFn()}</ToastContextProvider>
-)
 
 configure(loadStories, module);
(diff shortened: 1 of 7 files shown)


.storybook/config.js
```diff
@@ -7,10 +7,5 @@ const req = require.context("../stories", true, /\.stories\.tsx$/);
 function loadStories() {
   req.keys().forEach(req);
 }
- 
-const ToastDecorator = storyFn => (
-  <ToastContextProvider>{storyFn()}</ToastContextProvider>
-)
```
yasincaliskan, line 13: Is there any special reason to move `stories`?

yasincaliskan (approved):

.storybook/config.js
edizcelik: We don't need to add this decorator. I included multiple providers directly within the stories file

PR #99, opened 2021-05-24
fix: Add missing classnames to Toast and Input

yasincaliskan:
- Add `toast` container class
- Add `input-container--type-number` container class for `number input`

### DIFF
--- a/src/form/input/Input.tsx
+++ b/src/form/input/Input.tsx
@@ -83,11 +83,12 @@ function Input(props: InputProps) {
     setNumberSeparatorsForLocale
   ] = useState(() => getNumberSeparators(locale));
   const isNumberInput = type === "number";
-  const inputContainerClassName = classNames("input-container", customClassName);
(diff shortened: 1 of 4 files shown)


src/form/input/Input.tsx
```diff
@@ -83,11 +83,12 @@ function Input(props: InputProps) {
     setNumberSeparatorsForLocale
   ] = useState(() => getNumberSeparators(locale));
   const isNumberInput = type === "number";
-  const inputContainerClassName = classNames("input-container", customClassName);
+  const inputContainerClassName = classNames("input-container", customClassName, {
+    "input-container--type-number": isNumberInput
+  });
```
edizcelik, line 88:
```suggestion
  const inputContainerClassName = classNames("input-container", customClassName, `input-container--type-${type}`);
```

I think this is better and this way we would have added the modifier classname for all `type` props

edizcelik (changes requested):

edizcelik (approved):

PR #102, opened 2021-05-31
feat(button): Add `ref` prop

yasincaliskan:
- `ref` prop was added for can pass `ref` and use with `useRef` hook.
- It doesn't contain breaking change.
- It can be used like this;

```ts
const buttonRef = useRef<HTMLButtonElement | null>(null);

<Button ref={buttonRef} {...props} />
```

### DIFF
--- a/src/button/Button.tsx
+++ b/src/button/Button.tsx
@@ -8,7 +8,7 @@ import "./_button.scss";
 
 export type ButtonProps = Omit<
   React.ButtonHTMLAttributes<HTMLButtonElement>,
-  "disabled"
+  "disabled" | "className"
 > & {
   children: React.ReactNode;
   testid?: string;
--- a/src/button/Button.tsx
+++ b/src/button/Button.tsx
@@ -17,12 +17,6 @@ export type ButtonProps = Omit<
   shouldPreventDefault?: boolean;
   shouldStopPropagation?: boolean;
   shouldFocus?: boolean;
-  onClick?: React.ReactEventHandler<HTMLButtonElement>;
-  onMouseOver?: React.ReactEventHandler<HTMLButtonElement>;
-  onMouseDown?: React.ReactEventHandler<HTMLButtonElement>;
-  onMouseUp?: React.ReactEventHandler<HTMLButtonElement>;
-  onFocus?: React.ReactEventHandler<HTMLButtonElement>;
-  onBlur?: React.ReactEventHandler<HTMLButtonElement>;
   ref?: React.RefObject<HTMLButtonElement>;
   shouldDisplaySpinner?: boolean;
   ariaLabel?: string;
@@ -45,11 +39,6 @@ const Button = React.forwardRef<HTMLButtonElement, Record<string, any>>(
       shouldDisplaySpinner,
       ariaLabel,
       isDisabled,
-      onMouseOver,
-      onMouseDown,
-      onMouseUp,
-      onFocus,
-      onBlur,
       ...rest
     } = props;
     const isButtonDisabled = Boolean(isDisabled || shouldDisplaySpinner);
@@ -73,11 +62,6 @@ const Button = React.forwardRef<HTMLButtonElement, Record<string, any>>(
         // eslint-disable-next-line jsx-a11y/no-autofocus
         autoFocus={shouldFocus}
         onClick={handleClick}
-        onMouseOver={handleMouseOver}
-        onMouseDown={onMouseDown}
-        onMouseUp={onMouseUp}
-        onFocus={handleFocus}
-        onBlur={onBlur}
         disabled={isButtonDisabled}
         aria-label={ariaLabel}
         {...rest}>
@@ -87,7 +71,7 @@ const Button = React.forwardRef<HTMLButtonElement, Record<string, any>>(
       </button>
     );
 
-    function handleClick(event: React.SyntheticEvent<HTMLButtonElement>) {
+    function handleClick(event: React.MouseEvent<HTMLButtonElement>) {
       if (onClick) {
         if (shouldPreventDefault) {
           event.preventDefault();
@@ -100,18 +84,6 @@ const Button = React.forwardRef<HTMLButtonElement, Record<string, any>>(
         if (!isButtonDisabled) {
           onClick(event);
         }
-      }
-    }
-
-    function handleMouseOver(event: React.SyntheticEvent<HTMLButtonElement>) {
-      if (onMouseOver) {
-        onMouseOver(event);
-      }
-    }
-
-    function handleFocus(event: React.SyntheticEvent<HTMLButtonElement>) {
-      if (onFocus) {
-        onFocus(event);
       }
     }
   }
--- a/src/button/Button.tsx
+++ b/src/button/Button.tsx
@@ -17,6 +17,7 @@ export interface ButtonProps {
   onFocus?: React.ReactEventHandler<HTMLButtonElement>;
   onBlur?: React.ReactEventHandler<HTMLButtonElement>;
   id?: string;
+  ref?: React.RefObject<HTMLButtonElement>;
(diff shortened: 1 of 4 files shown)


src/button/Button.tsx
```diff
@@ -17,6 +17,7 @@ export interface ButtonProps {
   onFocus?: React.ReactEventHandler<HTMLButtonElement>;
   onBlur?: React.ReactEventHandler<HTMLButtonElement>;
   id?: string;
+  ref?: React.RefObject<HTMLButtonElement>;
```
edizcelik, line 20: let's use `React.ButtonHTMLAttributes<HTMLButtonElement>` when defining ButtonProps:
```
type ButtonProps = Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, "disabled"> & {
  children: React.ReactNode;
  testid?: string;
  customSpinner?: React.ReactNode;
  isDisabled?: boolean;
  shouldPreventDefault?: boolean; 
  shouldStopPropagation?: boolean;  
  shouldFocus?: boolean;
  /* other props that are missing from React.ButtonHTMLAttributes<HTMLButtonElement> */
}
```

src/button/Button.tsx
```diff
@@ -29,10 +30,18 @@ export interface ButtonProps {
...
 
+const RefButtonComponent = React.forwardRef<HTMLButtonElement, Record<string, any>>(
+  // eslint-disable-next-line prefer-arrow-callback
+  function RefButton(props, ref) {
+    return <button ref={ref} {...props} />;
+  }
+);
+
 function Button({
   testid,
```
edizcelik, line 41: You need to export the component created from forwardRef 
```suggestion
const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
  // eslint-disable-next-line prefer-arrow-callback
  function ButtonComponent(props, ref) {
    const {testid,  /* ... */ } = props;
    
    // ...
  }
);
```

edizcelik (changes requested):

src/button/Button.tsx
```diff
@@ -6,115 +6,124 @@ import Spinner from "../spinner/Spinner";
...
   lang?: string;
-  type?: "button" | "submit" | "reset";
   shouldDisplaySpinner?: boolean;
-  isDisabled?: boolean;
-  shouldPreventDefault?: boolean;
-  shouldStopPropagation?: boolean;
-  shouldFocus?: boolean;
   ariaLabel?: string;
   customClassName?: string;
   tabIndex?: number;
```
edizcelik, line 32: Please clean out the props that are already available from `React.ButtonHTMLAttributes<HTMLButtonElement>`. This was mentioned in my comment that proposed using `React.ButtonHTMLAttributes<HTMLButtonElement>`

src/button/Button.tsx
```diff
@@ -6,115 +6,124 @@ import Spinner from "../spinner/Spinner";
...
+      shouldFocus,
+      shouldDisplaySpinner,
+      ariaLabel,
+      isDisabled,
+      onMouseOver,
+      onMouseDown,
+      onMouseUp,
+      onFocus,
+      onBlur,
+      tabIndex
```
edizcelik, line 59: Let's destructure only the values that we actually need and gather the remaining props with a rest operator.
```suggestion
      /* props that we need to use */,
      ...otherProps
```

stories/3-Button.stories.tsx
```diff
@@ -6,106 +6,122 @@ import FileUploadButton from "../src/button/file-upload/FileUploadButton";
...
+          alert(`Button width: ${buttonRef.current.getBoundingClientRect().width}`)
+        }>
+        {"Get Button Width"}
+      </Button>
+
+      <br />
+      <hr />
+      <br />
+
+      <FileUploadButton
```
edizcelik, line 64: Can we create a story for FileUploadButton examples ?

edizcelik (changes requested):

stories/3-Button.stories.tsx
yasincaliskan: For use `ref` in the `FileUploadButton` as well, we should create with `forwardRef`. I don't think so we need this. But you think there is any use-case, we can add.

yasincaliskan: I'll add the functionality it needs, now. 👍🏻

src/button/Button.tsx
```diff
@@ -6,115 +6,115 @@ import Spinner from "../spinner/Spinner";
...
+  isDisabled?: boolean;
+  shouldPreventDefault?: boolean;
+  shouldStopPropagation?: boolean;
+  shouldFocus?: boolean;
   onClick?: React.ReactEventHandler<HTMLButtonElement>;
   onMouseOver?: React.ReactEventHandler<HTMLButtonElement>;
   onMouseDown?: React.ReactEventHandler<HTMLButtonElement>;
   onMouseUp?: React.ReactEventHandler<HTMLButtonElement>;
   onFocus?: React.ReactEventHandler<HTMLButtonElement>;
   onBlur?: React.ReactEventHandler<HTMLButtonElement>;
```
mucahit, line 25: I think we can also remove those props since it's extended from `React.ButtonHTMLAttributes<HTMLButtonElement>`

```suggestion
```

jamcry (approved):

bilgeyucel (approved):

src/button/Button.tsx
```diff
@@ -6,115 +6,88 @@ import Spinner from "../spinner/Spinner";
...
-  lang?: string;
-  type?: "button" | "submit" | "reset";
-  shouldDisplaySpinner?: boolean;
   isDisabled?: boolean;
   shouldPreventDefault?: boolean;
   shouldStopPropagation?: boolean;
   shouldFocus?: boolean;
+  ref?: React.RefObject<HTMLButtonElement>;
+  shouldDisplaySpinner?: boolean;
   ariaLabel?: string;
```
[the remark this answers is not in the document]
yasincaliskan: We can remove it. I'll add to release notes under the breaking changes title.

mucahit (approved):

Anlerkan (approved):

src/button/file-upload/FileUploadButton.tsx
```diff
@@ -10,6 +10,7 @@ export type FileUploadButtonProps = Omit<
   children: React.ReactNode;
   customClassName?: string;
   customLabelClassName?: string;
+  containerRef?: React.RefObject<HTMLDivElement>;
```
edizcelik, line 13: Let's use forwardRef here too

src/form/input/file/FileInput.tsx
```diff
@@ -18,57 +18,59 @@ export interface FileInputProps {
...
-}: FileInputProps) {
-  const containerClassName = classNames("file-input__container", customClassName);
-  const isInputDisabled = isPending || isDisabled;
-  const labelClassName = classNames("file-input__label", customLabelClassName, {
-    "file-input__label--is-disabled": isInputDisabled
-  });
-  const spinnerContent = customSpinner || (
-    <Spinner customClassName={"file-input__spinner"} />
-  );
+const FileInput = React.forwardRef<HTMLDivElement, Record<string, any>>(
```
edizcelik, line 24: I think it makes more sense for this component to forward this ref to label tag and remove labelRef prop

stories/3-Button.stories.tsx
edizcelik: I think we should separate regular button stories from file upload button examples

edizcelik (changes requested):

src/form/input/file/FileInput.tsx
edizcelik: Because we would probably need to access the label instead of the container

edizcelik (approved):

PR #103, opened 2021-06-06
feat(list): Add tests for List and ListItem

Anlerkan:
### Description

- Add test for `<List>` and `<ListItem>`

### Notes

- I could not test the `listItemKeyGenerator` prop on `<List>` component. I could not access `Fragment` because `key` is the only attribute that can be passed to `<Fragment>` . We can ignore this or maybe we can find a different way to render children.

### DIFF
--- a/src/list/item/ListItem.tsx
+++ b/src/list/item/ListItem.tsx
@@ -5,7 +5,7 @@ import classNames from "classnames";
 
 import {KEYBOARD_EVENT_KEY} from "../../core/utils/keyboard/keyboardEventConstants";
 
-interface ListItemProps {
+export interface ListItemProps {
   children?: React.ReactNode;
   testid?: string;
   customClassName?: string;
--- a/src/list/item/list-item.test.tsx
+++ b/src/list/item/list-item.test.tsx
@@ -1,0 +1,73 @@
+import React from "react";
+import {render, cleanup, fireEvent} from "@testing-library/react";
+import "@testing-library/jest-dom";
+import {create} from "react-test-renderer";
+
+import {testA11y} from "../../core/utils/test/testUtils";
+import ListItem, {ListItemProps} from "./ListItem";
+
+describe("<ListItem />", () => {
+  afterEach(cleanup);
+
+  const defaultListItemProps: ListItemProps = {
+    testid: "list-item",
+    children: <p data-testid={"list-item.content"}>{"Test"}</p>
+  };
+
+  it("should render correctly", () => {
+    render(<ListItem {...defaultListItemProps} />);
+  });
+
+  it("should matches snapshot", () => {
+    const tree = create(<ListItem {...defaultListItemProps} />).toJSON();
+
+    expect(tree).toMatchSnapshot();
+  });
+
+  it("should pass a11y test", async () => {
+    const {container} = render(<ListItem {...defaultListItemProps} />);
+
+    await testA11y(container, {rules: {listitem: {enabled: false}}});
+  });
+
+  it("should render children correctly", () => {
+    const {getByTestId} = render(<ListItem {...defaultListItemProps} />);
+
+    expect(getByTestId("list-item")).toContainElement(getByTestId("list-item.content"));
+  });
+
+  it("clickableListItemProps should work correctly", () => {
+    const handleClick = jest.fn();
+    const {getByTestId, getByRole} = render(
+      <ListItem
+        clickableListItemProps={{onClick: handleClick, tabIndex: 2}}
+        {...defaultListItemProps}
+      />
+    );
+    const listItemButton = getByRole("button");
+
+    fireEvent.click(listItemButton);
+
+    expect(getByTestId("list-item")).toContainElement(listItemButton);
+    expect(listItemButton).toHaveAttribute("tabIndex", "2");
+    expect(handleClick).toHaveBeenCalledTimes(1);
+  });
+
+  it("should run click event handler when enter key pressed", () => {
+    const handleClick = jest.fn();
+    const {getByRole} = render(
+      <ListItem
+        clickableListItemProps={{onClick: handleClick, tabIndex: 2}}
+        {...defaultListItemProps}
+      />
+    );
+
+    fireEvent.keyPress(getByRole("button"), {
+      key: "Enter",
+      code: "Enter",
+      keyCode: 13
+    });
+
+    expect(handleClick).toHaveBeenCalledTimes(1);
+  });
+});
--- a/src/list/list.test.tsx
+++ b/src/list/list.test.tsx
@@ -1,0 +1,102 @@
+import React from "react";
+import {render, cleanup, screen} from "@testing-library/react";
+import "@testing-library/jest-dom";
+import {create} from "react-test-renderer";
+
+import {testA11y} from "../core/utils/test/testUtils";
+import List, {ListProps} from "./List";
+import ListItem from "./item/ListItem";
+
+describe("<List />", () => {
+  afterEach(cleanup);
+
+  const listItems = ["Hipo", "Labs"];
+  const defaultListProps: ListProps = {
+    testid: "list",
+    items: listItems,
+    children: (item, testId, index) => (
+      <ListItem key={`list-item-${index}`} testid={testId}>
+        <span>{item}</span>
+      </ListItem>
+    )
+  };
+
+  it("should render correctly", () => {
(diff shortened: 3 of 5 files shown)


src/list/item/ListItem.tsx
```diff
@@ -5,7 +5,7 @@ import classNames from "classnames";
 
 import {KEYBOARD_EVENT_KEY} from "../../core/utils/keyboard/keyboardEventConstants";
 
-interface ListItemProps {
+export interface ListItemProps {
```
yasincaliskan, line 8: We can export this in `index.ts` instead.

```ts
import {ListItemProps as ListItemComponentProps} from "...";
...
export type ListItemProps = ListItemComponentProps;
```

Anlerkan: Thank you 🙏

src/list/list.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+import {render, cleanup, screen} from "@testing-library/react";
+import "@testing-library/jest-dom";
+import {create} from "react-test-renderer";
+
+import {testA11y} from "../core/utils/test/testUtils";
+import List, {ListProps} from "./List";
+import ListItem from "./item/ListItem";
+
+describe("<List />", () => {
+  afterEach(cleanup);
```
jamcry, line 11: Do we need to use `cleanup` (or, why do we need it)? In this [blog post](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library) it says :
> For a long time now cleanup happens automatically (supported for most major testing frameworks) and you no longer need to worry about it.

src/list/list.test.tsx
```diff
@@ -0,0 +1,116 @@
...
+        {...defaultListProps}
+      />
+    );
+
+    expect(getByTestId(defaultListProps.testid!)).toContainElement(
+      getByTestId("list.empty-state")
+    );
+  });
+
+  it("should render placeholder if both shouldDisplayPlaceholder and shouldDisplayEmptyState is true", () => {
```
jamcry, line 86:
```suggestion
  it("should render placeholder if both shouldDisplayPlaceholder and shouldDisplayEmptyState are true", () => {

```

src/list/item/list-item.test.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <ListItem
+        clickableListItemProps={{onClick: handleClick, tabIndex: 2}}
+        {...defaultListItemProps}
+      />
+    );
+
+    fireEvent.keyPress(getByRole("button"), {
+      key: "Enter",
+      code: "Enter",
+      keyCode: 13
```
jamcry, line 70: Do we have to pass all of them (key, code and keyCode)?

src/list/list.test.tsx
Anlerkan: Thank you. It seems we don't need to use cleanup. So i will remove it from all tests

src/list/item/list-item.test.tsx
Anlerkan: Yes you are right. Actually just keyCode is enough for here 🙏

mucahit (approved):

src/list/list.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+    expect(getByTestId(defaultListProps.testid!).children).toHaveLength(listItems.length);
+
+    for (let index = 0; index < listItems.length; index++) {
+      const listItemContent = getByTestId(`list.item-${index}`);
+
+      expect(getByTestId(defaultListProps.testid!).children[index]).toEqual(
+        listItemContent
+      );
+    }
```
edizcelik, line 49: This is heavily dependant on the implementation details. Please try to improve and keep it in line with the [guidelines](https://testing-library.com/docs/guiding-principles/) of testing-library

src/list/list.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+      <List
+        placeholderProps={{
+          shouldDisplayPlaceholder: true,
+          placeholder: <p data-testid={"list.placeholder"}>{"Loading..."}</p>
+        }}
+        {...defaultListProps}
+      />
+    );
+
+    expect(getByTestId(defaultListProps.testid!)).toContainElement(
```
edizcelik, line 63: Cant you use the container that is returned by the render function here ? Same goes for the other test blocks
```suggestion
    expect(container).toContainElement(

```

src/list/list.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+      getByTestId("list.empty-state")
+    );
+  });
+
+  it("should render placeholder if both shouldDisplayPlaceholder and shouldDisplayEmptyState are true", () => {
```
edizcelik, line 84:
```suggestion
  it("should render placeholder when shouldDisplayPlaceholder and shouldDisplayEmptyState are both true", () => {

```

But it seems to me that it should actually render the empty state in this case. I am not actually in favour of having this placeholderProps and emptyStateProps for List component, and would prefer to remove them altogether.

src/list/list.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+    expect(tree).toMatchSnapshot();
+  });
+
+  it("should pass a11y test", async () => {
+    const {container} = render(<List {...defaultListProps} />);
+
+    await testA11y(container);
+  });
+
+  it("should render children correctly", () => {
```
edizcelik, line 38:
```suggestion
  it("should render items correctly", () => {

```

src/list/list.test.tsx
```diff
@@ -0,0 +1,114 @@
...
+      getByTestId("list.placeholder")
+    );
+    expect(screen.queryByText("Not found.")).not.toBeInTheDocument();
+  });
+
+  it("should handle empty items correctly", () => {
+    const {getByTestId} = render(<List {...defaultListProps} items={[]} />);
+
+    expect(getByTestId(defaultListProps.testid!).children).toHaveLength(0);
+  });
```
edizcelik, line 113: I think this can be done within `should render items correctly` block

src/list/item/list-item.test.tsx
```diff
@@ -0,0 +1,71 @@
...
+      <ListItem
+        clickableListItemProps={{onClick: handleClick, tabIndex: 2}}
+        {...defaultListItemProps}
+      />
+    );
+    const listItemButton = getByRole("button");
+
+    fireEvent.click(listItemButton);
+
+    expect(getByTestId(defaultListItemProps.testid!)).toContainElement(listItemButton);
```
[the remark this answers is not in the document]
Anlerkan: Yes, the following tests ensure that this element exists, but with this test, I am verifying that this element is in `listitem` element. Is this test necessary? What do you think?

yasincaliskan (approved):

src/list/item/list-item.test.tsx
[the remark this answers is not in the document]
edizcelik: Ok that's fine we can keep this check as well but I think we need to re-order the checks here:

```
expect(getByTestId(defaultListItemProps.testid!)).toContainElement(listItemButton);
expect(listItemButton).toHaveAttribute("tabIndex", "2");

fireEvent.click(listItemButton);
expect(handleClick).toHaveBeenCalledTimes(1);
```

src/list/item/list-item.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+    expect(container).toContainElement(screen.getByText("Test"));
+  });
+
+  it("Can be clicked", () => {
+    const handleClick = jest.fn();
+
+    const {container} = render(
+      <ListItem
+        clickableListItemProps={{onClick: handleClick, tabIndex: 2}}
```
bilgeyucel, line 36: Why is the `tabIndex` `2`? Wouldn't it be enough if it was `0`?

Anlerkan: Default tabIndex is 0. It maybe not worth to test but I wanted to test that it also works when we give custom tabIndex
` tabIndex={clickableListItemProps.tabIndex || 0}`

edizcelik (approved):

src/list/list.test.tsx
Anlerkan: To avoid being dependent to implementation details, i added `listitem` role to list items so we can get them from their roles. We dont need testid's anymore. 
https://github.com/Hipo/react-ui-toolkit/pull/103/commits/b982cbe8868a164995e2a82138741d6a26052e2a

PR #104, opened 2021-06-06
feat(progress-bar): Add tests

Anlerkan:
### Description
- `<ProgressBar>` tests added.

### DIFF
--- a/src/progress-bar/ProgressBar.tsx
+++ b/src/progress-bar/ProgressBar.tsx
@@ -12,6 +12,7 @@ export interface ProgressBarProps {
   };
   children?: React.ReactNode;
   customClassName?: string;
+  testid?: string;
 }
 
 function ProgressBar(props: ProgressBarProps) {
@@ -19,7 +20,8 @@ function ProgressBar(props: ProgressBarProps) {
     percentage,
     style: {trackColor = "blue", backgroundColor = "gray", completedColor = "green"},
     children,
-    customClassName
+    customClassName,
+    testid
   } = props;
   const progressBarClassName = classNames("progress-bar", customClassName);
 
@@ -28,7 +30,8 @@ function ProgressBar(props: ProgressBarProps) {
       className={progressBarClassName}
       style={{
         backgroundColor
-      }}>
+      }}
+      data-testid={testid}>
       <div
         className={"progress-bar__track"}
         style={{
@@ -36,6 +39,7 @@ function ProgressBar(props: ProgressBarProps) {
           // eslint-disable-next-line no-magic-numbers
           backgroundColor: percentage === 100 ? completedColor : trackColor
         }}
+        data-testid={`${testid}.track`}
       />
 
       {children && <div className={"progress-bar__content"}>{children}</div>}
--- a/src/progress-bar/progress-bar.test.tsx
+++ b/src/progress-bar/progress-bar.test.tsx
@@ -1,0 +1,78 @@
+import React from "react";
+import {render, cleanup} from "@testing-library/react";
+import "@testing-library/jest-dom";
+import {create} from "react-test-renderer";
+import ProgressBar, {ProgressBarProps} from "./ProgressBar";
+import {testA11y} from "../core/utils/test/testUtils";
+
+describe("<ProgressBar />", () => {
+  afterEach(cleanup);
+
+  const defaultProgressBarProps: ProgressBarProps = {
+    testid: "progress-bar",
+    percentage: 57,
+    style: {backgroundColor: "blue", completedColor: "yellow", trackColor: "orange"}
+  };
+
+  it("should render correctly", () => {
+    render(<ProgressBar {...defaultProgressBarProps} />);
+  });
+
+  it("should matches snapshot", () => {
+    const tree = create(<ProgressBar {...defaultProgressBarProps} />).toJSON();
+
+    expect(tree).toMatchSnapshot();
+  });
+
+  it("should pass a11y test", async () => {
+    const {container} = render(<ProgressBar {...defaultProgressBarProps} />);
+
+    await testA11y(container);
+  });
+
+  it("should render children correctly", () => {
+    const progressBarContent = <span data-testid={"progress-bar.content"}>{"Test"}</span>;
+
+    const {getByTestId} = render(
+      <ProgressBar {...defaultProgressBarProps}>{progressBarContent}</ProgressBar>
+    );
+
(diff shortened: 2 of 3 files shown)


jamcry (approved):
👏

src/progress-bar/ProgressBar.tsx
```diff
@@ -28,14 +30,16 @@ function ProgressBar(props: ProgressBarProps) {
...
         backgroundColor
-      }}>
+      }}
+      data-testid={testid}>
       <div
         className={"progress-bar__track"}
         style={{
           width: `${percentage}%`,
           // eslint-disable-next-line no-magic-numbers
           backgroundColor: percentage === 100 ? completedColor : trackColor
```
jamcry, line 40: This is not related to tests but I am wondering if we should handle `percentage > 100` case for this component, user can pass a percentage that is bigger than 100, and looks like this would cause overflow (track would be too long). Maybe we can talk in our meeting

yasincaliskan: You're right. We can change this condition like this;
```js
backgroundColor: percentage >= 100 ? completedColor : trackColor
```

src/progress-bar/progress-bar.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  it("if the percentage is equal to 100, background color of the progress bar track should be completedColor", () => {
+    const {getByTestId} = render(
+      <ProgressBar {...defaultProgressBarProps} percentage={100} />
+    );
+
+    expect(getByTestId("progress-bar.track")).toHaveStyle(
+      `background-color: ${defaultProgressBarProps.style.completedColor}`
+    );
+  });
```
yasincaliskan, line 77: As Cem mentioned, maybe we add can one more test for a percentage greater than 100. Generally, we expect the percentage from API but there might be a case to control on the client side.

yasincaliskan (approved):

src/progress-bar/ProgressBar.tsx
yasincaliskan: You're right. We can change like this;
```js
backgroundColor: percentage >= 100 ? completedColor : trackColor
```

mucahit (approved):

src/progress-bar/ProgressBar.tsx
```diff
@@ -28,14 +30,17 @@ function ProgressBar(props: ProgressBarProps) {
       className={progressBarClassName}
       style={{
         backgroundColor
-      }}>
+      }}
+      data-testid={testid}>
```
edizcelik, line 34: Let's investigate how we can turn this into a more accessible ProgressBar. Maybe using this role helps and then we can get the element byRole in the tests.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_progressbar_role

src/progress-bar/progress-bar.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+describe("<ProgressBar />", () => {
+  const defaultProgressBarProps: ProgressBarProps = {
+    testid: "progress-bar",
+    percentage: 57,
+    style: {backgroundColor: "blue", completedColor: "yellow", trackColor: "orange"}
+  };
+
+  it("should render correctly", () => {
+    render(<ProgressBar {...defaultProgressBarProps} />);
+  });
```
edizcelik, line 17: Do we really need to wrap this with an `it` ?
```suggestion
  render(<ProgressBar {...defaultProgressBarProps} />);
```

src/progress-bar/progress-bar.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const progressBarContent = <span data-testid={"progress-bar.content"}>{"Test"}</span>;
+
+    const {getByTestId} = render(
+      <ProgressBar {...defaultProgressBarProps}>{progressBarContent}</ProgressBar>
+    );
+
+    expect(getByTestId(defaultProgressBarProps.testid!)).toContainElement(
+      getByTestId("progress-bar.content")
+    );
+  });
```
edizcelik, line 41: Can't you test this with the Snapshot test above ?

src/progress-bar/progress-bar.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    expect(getByTestId(defaultProgressBarProps.testid!)).toHaveStyle(
+      `background-color: ${defaultProgressBarProps.style.backgroundColor}`
+    );
+  });
+
+  it("the width of the progress bar track should be `percentage` percent", () => {
+    const {getByTestId} = render(<ProgressBar {...defaultProgressBarProps} />);
+
+    expect(getByTestId(`${defaultProgressBarProps.testid!}.track`)).toHaveStyle(
+      `width: ${defaultProgressBarProps.percentage}%`
```
edizcelik, line 55: This seems to depend on the implementation. For example, the implementation details may change for the component but the visual output may still be the same. I wonder if we can decouple this from the implementation and test only the appearance of it without it relying on the implementation detail. May not be possible for this component but worth to ponder about.

src/progress-bar/progress-bar.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  it("if the percentage is greater or equal to 100,  width of the progress bar track should be 100%", () => {
+    const {getByTestId} = render(
+      <ProgressBar {...defaultProgressBarProps} percentage={120} />
+    );
+
+    expect(getByTestId(`${defaultProgressBarProps.testid!}.track`)).toHaveStyle(
+      "width: 100%"
+    );
+  });
```
edizcelik, line 91: I think this can be checked within the block `Progress bar track has the right size`

src/progress-bar/progress-bar.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  it("progress bar's background color should be backgroundColor", () => {
+    const {getByTestId} = render(<ProgressBar {...defaultProgressBarProps} />);
+
+    expect(getByTestId(defaultProgressBarProps.testid!)).toHaveStyle(
+      `background-color: ${defaultProgressBarProps.style.backgroundColor}`
+    );
+  });
+
+  it("the width of the progress bar track should be `percentage` percent", () => {
```
edizcelik, line 51:
```suggestion
  it("Progress bar track has the right size", () => {
```

bilgeyucel (approved):

edizcelik (approved):

Issue #106, opened 2021-06-28
TypeaheadInput: Make sure Input prop changes are always reflected on TypeaheadInput 


edizcelik:
I can't currently use TypeaheadInput with `localizationOptions` prop:

```
Property 'localizationOptions' does not exist on type 'IntrinsicAttributes & TypeaheadInputProps'.
```

We should change our approach here so that TypeaheadInput props a bit more dynamic and that any changes on Input component is available on TypeaheadInput as well.

PR #108, opened 2021-06-28
fix(typeahead-input): Extend TypeaheadInputProps with InputProps

yasincaliskan:
### Description
- `TypeaheadInputProps` was extended with `InputProps`.
- Duplicated props were removed.

### Related Issue
- https://github.com/Hipo/react-ui-toolkit/issues/106

### DIFF
--- a/src/form/input/typeahead/TypeaheadInput.tsx
+++ b/src/form/input/typeahead/TypeaheadInput.tsx
@@ -1,55 +1,31 @@
 import React, {useEffect} from "react";
 import classNames from "classnames";
 
-import Input, {InputTypes} from "../Input";
+import Input, {InputProps, InputTypes} from "../Input";
 import useDebounce from "../../../core/utils/hooks/debounce";
 
-export interface TypeaheadInputProps {
+export type TypeaheadInputProps = Omit<
+  InputProps,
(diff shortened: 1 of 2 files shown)


src/form/input/typeahead/TypeaheadInput.tsx
```diff
@@ -1,55 +1,31 @@
...
 import classNames from "classnames";
 
-import Input, {InputTypes} from "../Input";
+import Input, {InputProps, InputTypes} from "../Input";
 import useDebounce from "../../../core/utils/hooks/debounce";
 
-export interface TypeaheadInputProps {
+export type TypeaheadInputProps = Omit<
+  InputProps,
+  "onChange" | "type" | "customClassName" | "value"
```
edizcelik, line 9: I think we don't need to omit `customClassName` and `value` as they have the same type in InputProps as well

yasincaliskan: Yes, you are right. Just destructuring is enough.

edizcelik (approved):

mucahit (approved):

bilgeyucel (approved):

jamcry (approved):

Issue #107, opened 2021-06-28
Input: `has-error` and `is-disabled` classnames should be added to the container


edizcelik:
Currently they are just added to the Input. However, to make the component more customizable, we need to add these classes to the container. For example, a user of this component may want to change the color of an icon provided to the component (as leftIcon or rightIcon).

yasincaliskan:
Of course! So, should it be `input-container--has-error` and `input-container--is-disabled`?

edizcelik:
Yes.

PR #110, opened 2021-06-28
fix(input): Move `is-disabled` & `has-error` classNames to container

yasincaliskan:
### Breaking Change
- If `input--has-error` or `input--is-disabled` classNames are used in any project, these classNames should update with `input-container--has-error` and `input-container--is-disabled`.

### Related Ticket
- https://github.com/Hipo/react-ui-toolkit/issues/107

### DIFF
--- a/src/form/input/Input.tsx
+++ b/src/form/input/Input.tsx
@@ -97,12 +97,12 @@ function Input(props: InputProps) {
   const inputContainerClassName = classNames(
     "input-container",
     customClassName,
-    `input-container--type-${type}`
+    `input-container--type-${type}`,
+    {
+      "input-container--is-disabled": isDisabled,
+      "input-container--has-error": hasError
+    }
   );
-  const inputClassName = classNames("input", {
-    "input--is-disabled": isDisabled,
-    "input--has-error": hasError
-  });
   let finalValue = value;
 
   if (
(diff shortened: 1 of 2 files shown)


src/form/input/Input.tsx
```diff
@@ -97,12 +97,12 @@ function Input(props: InputProps) {
...
+    `input-container--type-${type}`,
+    {
+      "input-container--is-disabled": isDisabled,
+      "input-container--has-error": hasError
+    }
   );
-  const inputClassName = classNames("input", {
-    "input--is-disabled": isDisabled,
-    "input--has-error": hasError
-  });
```
edizcelik, line 105: Let's make sure we explain this in the release notes as this is a breaking change

src/form/input/_input.scss
```diff
@@ -37,6 +37,6 @@
   }
 }
 
-.input--has-error {
+.input-container--has-error {
   border: 1px solid var(--invalid-text-color);
 }
```
edizcelik, line 42: Do we want to give the border to the container or the input ? I guess adding it to the container is fine but let's make sure it is consistent

yasincaliskan: I don't think it will be a problem. It works fine.
image

edizcelik (approved):

Anlerkan (approved):

mucahit (approved):
This will be a breaking change

bilgeyucel (approved):

jamcry (approved):

Issue #109, opened 2021-06-28
Input: Strange value when using Input with type="number"


edizcelik:
Input component that I am using looks like this:
```
<Input
              type={"number"}
              name={"custom-slippage-tolerance"}
              placeholder={"0.50%"}
              customClassName={"settings-modal__slippage-tolerance-row__input"}
              onChange={handleCustomSlippageTolerance}
              rightIcon={
                slippageToleranceComponent === "input" &&
                state.slippageTolerancePercentage
                  ? "%"
                  : null
              }
              value={
                slippageToleranceComponent === "input"
                  ? state.slippageTolerancePercentage
                  : ""
              }
              localizationOptions={{
                maximumFractionDigits: 3,
                shouldFormatToLocaleString: true
              }}
/>
```

When I type "1.5" to the input and then remove "1", `event.currentTarget.value` within onChange callback looks like this: `NaN.5`. However, this should have been ".5" instead.
<img width="714" alt="Screen Shot 2021-06-28 at 6 56 26 PM" src="https://user-images.githubusercontent.com/22460300/123668069-68f18800-d843-11eb-9ebf-ebcd7bb1b1db.png">

PR #111, opened 2021-06-29
fix(input): Don't parse integer if integer part is empty

yasincaliskan:
### Description
- When deleted a single integer character, the integer part equals `empty string` and `parseInt(<empty string>)` equals `NaN`.
- With this condition, `integerPart` will not be `NaN`. `0.5 -> .5`

Fixes: https://github.com/Hipo/react-ui-toolkit/issues/109

### DIFF
--- a/src/core/utils/number/numberUtils.ts
+++ b/src/core/utils/number/numberUtils.ts
@@ -156,7 +156,8 @@ function removeLeadingZeros(locale = navigator.language, value: string) {
     integerPart.length !== String(parseInt(integerPart)).length &&
     integerPart !== LOCALE_MINUS_SIGN &&
     integerPart !== LOCALE_NEGATIVE_ZERO &&
-    integerPart !== DEFAULT_NEGATIVE_ZERO
+    integerPart !== DEFAULT_NEGATIVE_ZERO &&
+    integerPart !== ""
   ) {
     integerPart = String(parseInt(integerPart));
   }


src/core/utils/number/numberUtils.ts
```diff
@@ -156,7 +156,8 @@ function removeLeadingZeros(locale = navigator.language, value: string) {
     integerPart.length !== String(parseInt(integerPart)).length &&
     integerPart !== LOCALE_MINUS_SIGN &&
     integerPart !== LOCALE_NEGATIVE_ZERO &&
-    integerPart !== DEFAULT_NEGATIVE_ZERO
+    integerPart !== DEFAULT_NEGATIVE_ZERO &&
+    integerPart !== ""
```
edizcelik, line 160: Should we do this instead ?

```
    integerPart &&
    integerPart.length !== String(parseInt(integerPart)).length &&
    integerPart !== LOCALE_MINUS_SIGN &&
    integerPart !== LOCALE_NEGATIVE_ZERO &&
    integerPart !== DEFAULT_NEGATIVE_ZERO
```

yasincaliskan: I think we shouldn't. Because if you delete `integerPart` completely, `String(parseInt(integerPart))` will be `NaN` and they wouldn't be equal. So, this check wouldn't work.
```js
String(parseInt(integerPart)).length = 3
"NaN".length = 3
```

edizcelik: I am not sure I understand. When we remove the integer part of the value, `integerPart` variable would be equal to empty string, `""`, is it not?

yasincaliskan: Actually not. Let's think step by step.
If we delete `0` from `0.5`,
- `integerPart` = `""` (empty string)
- `parseInt("")` = `"NaN"`
- `String(NaN).length` = 3

image
image

edizcelik: So `integerPart && integerPart.length !== String(parseInt(integerPart)).length` would equal to empty string, `""` and therefore the whole check would be coerced as a false value

yasincaliskan: You're right! Adding `integerPart` as a condition, makes sense. It works!

edizcelik (approved):

mucahit (approved):

Anlerkan (approved):

bilgeyucel (approved):

jamcry (approved):

PR #121, opened 2021-07-06
feat(tab): Add onTabChange prop

### LINKED ISSUE #115: Tab: Add `onTabChange` Prop
We need to add `onTabChange` prop to the Tab component to listen to tab changes.

Pass target (a.k.a the active item index) `index` to `onTabChange`

```diff
export interface TabProps {
  items: TabItem[];
  children: React.ReactNode[];
  testid?: string;
  initialActiveTabIndex?: number;
  customClassName?: string;
+ onTabChange: (index: number) => void;
}
```

mfarukkoc:
### Description
- Added optional `onTabChange` prop to the Tab component that triggers on tab change and takes index as parameter.
- Added example usage on Tab story page.
```diff
export interface TabProps {
  items: TabItem[];
  children: React.ReactNode[];
  testid?: string;
  initialActiveTabIndex?: number;
  customClassName?: string;
+ onTabChange?: (index: number) => void;
}
```
Closes #115

### DIFF
--- a/src/tab/Tab.tsx
+++ b/src/tab/Tab.tsx
@@ -18,6 +18,7 @@ export interface TabProps {
   testid?: string;
   initialActiveTabIndex?: number;
   customClassName?: string;
+  onTabChange?: (index: number) => void;
 }
 
 function Tab({
@@ -25,7 +26,8 @@ function Tab({
   items,
   initialActiveTabIndex = 0,
   children,
-  customClassName
+  customClassName,
+  onTabChange
 }: TabProps) {
   const [activeTabIndex, setActiveTabIndex] = useState(initialActiveTabIndex);
   const tabClassName = classNames("tab", customClassName);
@@ -50,6 +52,8 @@ function Tab({
 
   function handleChangeActiveTab(index: number) {
     setActiveTabIndex(index);
+
+    if (onTabChange) onTabChange(index);
   }
 }
 
--- a/stories/6-Tab.stories.tsx
+++ b/stories/6-Tab.stories.tsx
@@ -16,7 +16,11 @@ const tabItems: TabItem[] = [
 
 storiesOf("Tab", module).add("Tab", () => (
   <Fragment>
-    <Tab items={tabItems}>
+    <Tab
+      items={tabItems}
+      onTabChange={(index) => {
+        console.log("tab changed to index: ", index);
+      }}>
       {[<div key={0}>{"Home"}</div>, <div key={1}>{"Following"}</div>]}
     </Tab>
   </Fragment>


src/tab/Tab.tsx
```diff
@@ -50,6 +52,8 @@ function Tab({
 
   function handleChangeActiveTab(index: number) {
     setActiveTabIndex(index);
+
+    if (onTabChange) onTabChange(index);
```
mucahit, line 56:
```suggestion
    if (onTabChange) {
      onTabChange(index);
    }
```

mucahit (changes requested):

yasincaliskan (approved):

mucahit:
As we discussed, let's tackle these two points;
- Let's add a prop that takes an index called `selectedItem` to make the Tab component controlled
- Let's remove `initialActiveTabIndex `

src/tab/Tab.tsx
```diff
@@ -16,20 +16,26 @@ export interface TabProps {
   items: TabItem[];
   children: React.ReactNode[];
   testid?: string;
-  initialActiveTabIndex?: number;
+  selectedItem?: number;
```
jamcry, line 19:
```suggestion
  activeTabIndex?: number;
```
I think this would be more descriptive, because we are passing the index rather than the item itself.

This causes duplicate names, you can solve it by renaming this while destructring in line 27:
```
activeTabIndex: activeTabIndexFromProps = 0
```

src/tab/Tab.tsx
```diff
@@ -16,20 +16,26 @@ export interface TabProps {
...
+  customClassName,
+  onTabChange
 }: TabProps) {
-  const [activeTabIndex, setActiveTabIndex] = useState(initialActiveTabIndex);
+  const [activeTabIndex, setActiveTabIndex] = useState(selectedItem);
   const tabClassName = classNames("tab", customClassName);
 
+  useEffect(() => {
+    setActiveTabIndex(selectedItem);
+  }, [selectedItem]);
```
jamcry, line 37: I think this is OK for now, but I don't think a component being both controlled and uncontrolled is a very good idea. Let's discuss this in the next team meeting and finalize the decision.

src/tab/Tab.tsx
```diff
@@ -16,20 +16,26 @@ export interface TabProps {
   items: TabItem[];
   children: React.ReactNode[];
   testid?: string;
-  initialActiveTabIndex?: number;
+  activeTabIndex?: number;
```
edizcelik, line 19: There should be two props: `activeTabIndex` and `initialActiveTabIndex`. If `activeTabIndex` is provided, then that should mean Tab component shouldn't handle the state itself but to delegate state management to its parent (ie become a controlled component). `initialActiveTabIndex` is there to set the active initial tab when  `activeTabIndex` was not provided.

See https://reactjs.org/docs/uncontrolled-components.html

src/tab/Tab.tsx
```diff
@@ -16,20 +16,26 @@ export interface TabProps {
...
+  customClassName,
+  onTabChange
 }: TabProps) {
-  const [activeTabIndex, setActiveTabIndex] = useState(initialActiveTabIndex);
+  const [activeTabIndex, setActiveTabIndex] = useState(activeTabIndexFromProps);
   const tabClassName = classNames("tab", customClassName);
 
+  useEffect(() => {
+    setActiveTabIndex(activeTabIndexFromProps);
+  }, [activeTabIndexFromProps]);
```
edizcelik, line 37: If we change the implementation to satisfy the requirements I described in my previous comment, I think we will be able to remove this useEffect block altogether.

src/tab/Tab.tsx
```diff
@@ -49,6 +55,9 @@ function Tab({
   );
 
   function handleChangeActiveTab(index: number) {
+    if (onTabChange && index !== activeTabIndex) {
+      onTabChange(index);
+    }
     setActiveTabIndex(index);
```
edizcelik, line 61: There shouldn't be a state duplication. If it is a controlled component, Tab's own state should never change but the component will be controlled via the prop passed.

stories/6-Tab.stories.tsx
```diff
@@ -16,8 +18,41 @@ const tabItems: TabItem[] = [
 
 storiesOf("Tab", module).add("Tab", () => (
   <Fragment>
-    <Tab items={tabItems}>
+    <p>Uncontrolled Tab</p>
+    <Tab
+      items={tabItems}
+      onTabChange={(index) => {
```
edizcelik, line 24: It should always get the `activeTabIndex` prop when there is a `onTabChange` because that means we want to control it.

edizcelik (changes requested):

stories/6-Tab.stories.tsx
mfarukkoc: I updated the TabProps to fix mistakes like that. When one of the controlled props (`onTabChange` or `activeTabIndex`) is provided other is required too, and `initialActiveTabIndex` shouldn't be defined.

mucahit (approved):

PR #123, opened 2021-07-07
feat(toast): Add limit prop to ToastProvider

### LINKED ISSUE #112: Toast: Add a limit to the number of toasts that can be visible at one time
Currently we don't have any limit on the number of toasts that can be added to the state. However, there is a requirement I am facing at the moment which states that there should be at most 5 toasts visible on screen at a given time. I believe we can control this behaviour by providing a prop to the ToastProvider and setting up the Toasts for a particular project. By default, we wouldn't add a limit to the number of toasts, but this way we would give a way to control this on the project level.

mfarukkoc:
### Description

- Added optional limit prop to limit the maximum number of toasts.

Closes #112


src/toast/util/toastTypes.ts
```diff
@@ -10,6 +10,7 @@ export type ToastAction =
   | {
       type: "DISPLAY";
       toastData: Omit<ToastData, "id"> & {id: string};
+      limit?: number;
```
edizcelik, line 13: We shouldn't need to provide this DISPLAY action with the limit. Let's find another solution which doesn't require passing the limit to the action.

src/toast/util/toastReducer.ts
```diff
@@ -11,14 +11,21 @@ function toastReducer(state: ToastState, action: ToastAction): ToastState {
...
     case "DISPLAY": {
-      const {toastData} = action;
+      const {toastData, limit} = action;
+
+      const toastStack = [
+        ...state.toastStack.filter(not(isSameToast(toastData.id))),
+        toastData
+      ];
+
+      if (limit)
```
edizcelik, line 21: Let's always use curly braces when defining if blocks

src/toast/util/toastReducer.ts
```diff
@@ -11,14 +11,21 @@ function toastReducer(state: ToastState, action: ToastAction): ToastState {
...
+        if (toastStack.length > limit)
```
edizcelik, line 22: Let's combine these two if statements so we have just one block

edizcelik (changes requested):

src/toast/ToastProvider.tsx
```diff
@@ -26,11 +27,13 @@ interface ToastContextProviderProps {
...
-  autoCloseToasts = true
+  autoCloseToasts = true,
+  limit
 }: ToastContextProviderProps) {
   const [state, dispatch] = useReducer(toastReducer, {
     ...initialToastState,
-    autoCloseToasts
+    autoCloseToasts,
+    limit
   });
```
edizcelik, line 38: Can you add an effect here that listens the changes on `limit` prop and reflects the changes in `limit` within the state later on ? Same thing for the `autoCloseToasts` prop. The users of the toast provider may have a need to change the limit or the autoCloseToasts dynamically in the lifetime of an app, e.g. limit may change when the screen size is small.

src/toast/util/toastReducer.ts
```diff
@@ -13,12 +13,21 @@ function toastReducer(state: ToastState, action: ToastAction): ToastState {
     case "DISPLAY": {
       const {toastData} = action;
 
+      const {limit} = state;
+
+      const toastStack = [
```
edizcelik, line 18:
```suggestion
      const {toastData} = action;
      const {limit} = state;
      const toastStack = [
```

edizcelik (changes requested):

src/toast/util/toastReducer.ts
```diff
@@ -12,13 +12,20 @@ function toastReducer(state: ToastState, action: ToastAction): ToastState {
...
       const {toastData} = action;
+      const {limit} = state;
+      const toastStack = [
+        ...state.toastStack.filter(not(isSameToast(toastData.id))),
+        toastData
+      ];
+
+      if (limit && toastStack.length > limit) {
+        // pop first toast if stack exceeds limit
+        toastStack.shift();
```
mucahit, line 23: I'd prefer to do it with `slice` because `shift` directly mutates the `toastStack`

mucahit (changes requested):
A minor request 👇

mucahit (approved):

src/toast/util/toastReducer.ts
```diff
@@ -57,6 +64,29 @@ function toastReducer(state: ToastState, action: ToastAction): ToastState {
       break;
     }
 
+    case "SET_LIMIT": {
+      const {limit} = action;
+      // prune the toasts exceeding the limit
+      const toastStack = state.toastStack.slice(
+        Math.max(0, state.toastStack.length - limit)
+      );
```
jamcry, line 72:
```suggestion
      if (limit && toastStack.length > limit) {
        // prune the toasts exceeding the limit
        toastStack = toastStack.slice(toastStack.length - limit);
      }
```

We can use the approach you used in "DISPLAY", I think that is more expressive

src/toast/util/toastReducer.ts
```diff
@@ -57,6 +64,29 @@ function toastReducer(state: ToastState, action: ToastAction): ToastState {
...
+    case "SET_AUTO_CLOSE": {
+      const {autoCloseToasts} = action;
+
+      newState = {
+        ...state,
+        autoCloseToasts
+      };
+      break;
+    }
     default:
```
jamcry, line 90:
```suggestion
      break;
    }
    
    case "SET_AUTO_CLOSE": {
      const {autoCloseToasts} = action;

      newState = {
        ...state,
        autoCloseToasts
      };
      break;
    }
    
    default:
```

Let's give them some space

src/toast/util/toastReducer.ts
```diff
@@ -57,6 +64,29 @@ function toastReducer(state: ToastState, action: ToastAction): ToastState {
...
+        limit
+      };
+      break;
+    }
+    case "SET_AUTO_CLOSE": {
+      const {autoCloseToasts} = action;
+
+      newState = {
+        ...state,
+        autoCloseToasts
```
jamcry, line 86:
```suggestion
      newState = {
        ...state,
        autoCloseToasts: action.autoCloseToasts
```
It is not used anywhere else, so we can directly use it inside

src/toast/ToastProvider.tsx
jamcry: @mfarukkoc Looks like you made the changes Ediz mentioned. After making changes mentioned in a comment, you can "Resolve conversation" using the button below.

yasincaliskan (approved):
Looks good, thank you!

src/toast/ToastProvider.tsx
```diff
@@ -26,13 +27,25 @@ interface ToastContextProviderProps {
...
 
+  useEffect(() => {
+    if (limit) {
```
edizcelik, line 40: Kind of an extreme edge case but what if they provide limit as 0 ?

mfarukkoc: Updated this statement with `limit !== undefined`

mucahit (approved):

PR #126, opened 2021-07-07
refactor(radioInput):  BEMify classNames

### LINKED ISSUE #125: RadioInput: BEMify classNames
Let's make the `radio-input-label` className BEMified

```diff
  radio-input-label
+ radio-input__label

  radio-input-label__icon
+ radio-input__label__icon
```

mfarukkoc:
### Description

- Changed 'radio-input-label' className to 'radio-input__label'.

Closes #125

### DIFF
--- a/src/form/input/radio/RadioInput.tsx
+++ b/src/form/input/radio/RadioInput.tsx
@@ -35,9 +35,9 @@ function RadioInput({testid, item, onSelect, isSelected, isDisabled}: RadioInput
     customClassName,
     content
   } = item;
-  const containerClassName = classNames("radio-input-label", customClassName, {
-    "radio-input-label--is-selected": isSelected,
-    "radio-input-label--is-disabled": isDisabled
+  const containerClassName = classNames("radio-input__label", customClassName, {
+    "radio-input__label--is-selected": isSelected,
+    "radio-input__label--is-disabled": isDisabled
   });
 
   return (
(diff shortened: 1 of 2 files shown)


src/form/input/radio/RadioInput.tsx
```diff
@@ -35,9 +35,9 @@ function RadioInput({testid, item, onSelect, isSelected, isDisabled}: RadioInput
     customClassName,
     content
   } = item;
-  const containerClassName = classNames("radio-input-label", customClassName, {
-    "radio-input-label--is-selected": isSelected,
-    "radio-input-label--is-disabled": isDisabled
+  const containerClassName = classNames("radio-input__label", customClassName, {
+    "radio-input__label--is-selected": isSelected,
+    "radio-input__label--is-disabled": isDisabled
```
edizcelik, line 40: This is misleading and inconsistent with the BEM convention. When we have `radio-input__label`, it means we have a `radio-input` block and `radio-input__label` element. Blocks should always wrap the elements in hierarchy. But when looking at the DOM structure for this component `radio-input` is under `radio-input__label` so it is a bit confusing. See https://github.com/Hipo/web-handbook/blob/master/css-guideline.md#naming-your-selectors.

mucahit: You're right, my bad. Let's close #126 and #127 @mfarukkoc

mfarukkoc:
Changes do not fit the BEM convention.

PR #128, opened 2021-07-07
feat(typeahead): Change input style

### LINKED ISSUE #45: TypeaheadSelect: Input should stretch dynamically and come after tags
Screen Shot 2020-12-11 at 3 45 12 PM

mfarukkoc:
### Description

- Changed style of the Typeahead input to stretch input dynamically and fit after tags.
- Increased the width of the Typeahead story page to show the dynamic stretching better.
- Added `extra` prop to the List to render an extra component after list items, input should be in the same flex-box with the tags for the dynamic adjustment when tags overflow to a new line.
- Renamed `.typeahead-select__tag-list` to `.typeahead-select__header-list` because list is no longer just rendering tags but also renders input.

Closes #45


src/select/typeahead/list/typeaheadList.tsx
```diff
@@ -0,0 +1,43 @@
...
+  customClassName?: string;
+  input?: React.ReactNode;
+}
+
+function TypeaheadList({
+  tags,
+  customClassName,
+  handleTagRemove,
+  input
+}: TypeaheadListProps) {
```
yasincaliskan, line 21: I'm wondering; Why do we need a separated `TypeaheadList` component? Will we be able to use this component as a single?

src/select/typeahead/TypeaheadSelect.tsx
```diff
@@ -113,43 +112,31 @@ function TypeaheadSelect({
...
-          onQueryChange={handleKeywordChange}
-          onKeyDown={handleKeyDown}
-          rightIcon={
-            areOptionsFetching ? spinnerContent : <CaretDownIcon aria-hidden={true} />
-          }
-          onFocus={handleTypeaheadInputFocus}
-          isDisabled={isDisabled}
-        />
-      )}
+      <TypeaheadList
```
mucahit, line 115: Let's rename it

```suggestion
      <TypeaheadSelectHeader
```

src/select/typeahead/list/typeaheadList.tsx
mucahit: We needed to put the TypeaheadInput into the tag list to improve UX but we had to separate it from the List component. Probably we're not gonna use it outside of the Typeahead.

mucahit (changes requested):

src/select/typeahead/header/TypeaheadSelectHeader.tsx
```diff
@@ -0,0 +1,43 @@
...
+  input?: React.ReactNode;
+}
+
+function TypeaheadSelectHeader({
+  tags,
+  customClassName,
+  handleTagRemove,
+  input
+}: TypeaheadSelectHeaderProps) {
+  const listClassName = classNames("typeahead-select__header-list", customClassName);
```
jamcry, line 22:
```suggestion
  const listClassName = classNames("typeahead-select-header__tag-list", customClassName);
```
Since it is a list of tags, I think this makes more sense.

Also, `TypeaheadSelectHeader` is a separate component, so `typeahead-select-header` looks more suitable than `typeahead-select__header`

src/select/typeahead/header/TypeaheadSelectHeader.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <ul className={listClassName}>
+      {tags.map((tag: TagShape) => (
+        <Fragment key={tag.id}>
+          <ListItem customClassName={"typeahead-select__header-list__item"}>
+            <Tag
+              onRemove={handleTagRemove}
+              customClassName={"typeahead-select__tag"}
```
jamcry, line 31: This would be
```suggestion
              customClassName={"typeahead-select-header__tag-list__item__tag"}
```

src/select/typeahead/header/TypeaheadSelectHeader.tsx
```diff
@@ -0,0 +1,43 @@
...
+function TypeaheadSelectHeader({
+  tags,
+  customClassName,
+  handleTagRemove,
+  input
+}: TypeaheadSelectHeaderProps) {
+  const listClassName = classNames("typeahead-select__header-list", customClassName);
+
+  return (
+    <ul className={listClassName}>
```
jamcry, line 25:
```suggestion
    <ul className={classNames("typeahead-select-header__tag-list", customClassName);}>
```
We can omit the variable to make the file more clean and simple

src/select/typeahead/header/TypeaheadSelectHeader.tsx
mfarukkoc: I'm not sure about `tag-list`,  since this list also contains our Input component.

src/select/typeahead/list/typeaheadList.tsx
yasincaliskan: I get it, thank you!

src/select/typeahead/header/TypeaheadSelectHeader.tsx
```diff
@@ -0,0 +1,43 @@
...
+      {tags.map((tag: TagShape) => (
+        <Fragment key={tag.id}>
+          <ListItem customClassName={"typeahead-select__header-list__item"}>
```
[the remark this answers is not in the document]
yasincaliskan: Of course. I don't think we need to use Fragment.

src/select/typeahead/header/TypeaheadSelectHeader.tsx
```diff
@@ -0,0 +1,41 @@
...
+          customClassName={"typeahead-select-header__tag-list__item"}>
+          <Tag
+            onRemove={handleTagRemove}
+            customClassName={"typeahead-select-header__tag-list__item__tag"}
+            tag={tag}
+          />
+        </ListItem>
+      ))}
+
+      {input}
```
edizcelik, line 36: We should probably wrap this with a `li` tag

edizcelik (approved):

mucahit (approved):

PR #130, opened 2021-07-08
feat(typeahead): Focus to input after option is selected

### LINKED ISSUE #46: TypeaheadSelect: Focus to input after an option is selected if it is a multi-select typeahead

mfarukkoc:
### Description
- Change inputRef target to the input instead of the inputContainer.
- Set ShouldFocusOnInput true after option selection.
- Input disappears after the selected options limit is reached, so it will not focus after that.
Closes #46

### DIFF
--- a/src/form/input/Input.tsx
+++ b/src/form/input/Input.tsx
@@ -56,6 +56,7 @@ export type InputProps = Omit<
     locale?: string;
     maximumFractionDigits?: number;
   };
+  inputRef?: React.RefObject<HTMLInputElement>;
 };
 
 /* eslint-disable complexity */
@@ -75,6 +76,7 @@ function Input(props: InputProps) {
     autoCorrect = "off",
     inputContainerRef,
     onChange,
+    inputRef,
     ...rest
   } = props;
   const {
@@ -161,6 +163,7 @@ function Input(props: InputProps) {
       )}
 
       <input
+        ref={inputRef}
         className={inputClassName}
         type={isNumberInput ? "text" : type}
         autoComplete={autoComplete}
--- a/src/form/input/typeahead/TypeaheadInput.tsx
+++ b/src/form/input/typeahead/TypeaheadInput.tsx
@@ -24,6 +24,7 @@ export interface TypeaheadInputProps {
   role?: string;
   children?: React.ReactNode;
   inputContainerRef?: React.RefObject<HTMLDivElement>;
+  inputRef?: React.RefObject<HTMLInputElement>;
 }
 
 const DEFAULT_DEBOUNCE_TIMEOUT = 250;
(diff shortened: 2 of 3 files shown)


src/form/input/Input.tsx
```diff
@@ -56,6 +56,7 @@ export type InputProps = Omit<
     locale?: string;
     maximumFractionDigits?: number;
   };
+  inputRef?: React.RefObject<HTMLInputElement>;
```
edizcelik, line 59: Let's use `forwardRef` instead. This will allow this component to take the ref that it receives as a prop and then we would forward this ref to the `input` element within this component.

src/form/input/typeahead/TypeaheadInput.tsx
```diff
@@ -24,6 +24,7 @@ export interface TypeaheadInputProps {
   role?: string;
   children?: React.ReactNode;
   inputContainerRef?: React.RefObject<HTMLDivElement>;
+  inputRef?: React.RefObject<HTMLInputElement>;
```
edizcelik, line 27: Same here. Let's use `forwardRef` for this component as well.

src/select/typeahead/TypeaheadSelect.tsx
```diff
@@ -135,7 +137,8 @@ function TypeaheadSelect({
         <TypeaheadInput
           testid={`${testid}.search`}
           customClassName={"typeahead-select__input"}
-          inputContainerRef={typeaheadInputRef}
+          inputContainerRef={typeaheadInputContainerRef}
```
edizcelik, line 140: It seems we don't use the ref for the container any longer. So we can remove it.

src/form/input/Input.tsx
```diff
@@ -58,195 +57,201 @@ export type InputProps = Omit<
...
-  }
-
-  if (isNumberInput && typeof value === "string" && shouldFormatToLocaleString) {
-    const [integerPart, decimalPart] = String(value).split(".");
-    const numberFormatter = formatNumber({
-      maximumFractionDigits,
-      locale
+const Input = React.forwardRef<HTMLInputElement, InputProps>(
+  /* eslint-disable complexity */
+  (props: InputProps, ref) => {
```
edizcelik, line 62: I think TSC can infer the type for props from by the definition of `React.forwardRef<HTMLInputElement, InputProps>`
```suggestion
  (props, ref) => {
```

src/form/input/Input.tsx
```diff
@@ -49,7 +49,6 @@ export type InputProps = Omit<
   isDisabled?: boolean;
   hasError?: boolean;
   customClassName?: string;
-  inputContainerRef?: React.RefObject<HTMLDivElement>;
```
edizcelik, line 52: This will be a breaking change. Let's make sure to include it in the release notes. cc: @mucahit

src/form/input/typeahead/TypeaheadInput.tsx
```diff
@@ -23,69 +23,69 @@ export interface TypeaheadInputProps {
   rightIcon?: React.ReactNode;
   role?: string;
   children?: React.ReactNode;
-  inputContainerRef?: React.RefObject<HTMLDivElement>;
```
edizcelik, line 26: same here. Let's make sure to note this change on the release note

PR #136, opened 2021-07-22
feat(switch): Add tests

Anlerkan:
### Description

* `<Switch />` tests implementation
* Accessibility improvements

### DIFF
--- a/src/switch/switch.test.tsx
+++ b/src/switch/switch.test.tsx
@@ -1,0 +1,60 @@
+import React from "react";
+import {render, screen} from "@testing-library/react";
+import "@testing-library/jest-dom";
+import userEvent from "@testing-library/user-event";
+
+import {testA11y} from "../core/utils/test/testUtils";
+import Switch, {SwitchProps} from "./Switch";
+
+describe("<Switch />", () => {
+  const defaultSwitchProps: SwitchProps = {
+    isToggledOn: true,
+    onToggle: jest.fn(),
+    testid: "switch"
+  };
+
+  it("should render correctly", () => {
+    render(<Switch {...defaultSwitchProps} />);
+  });
+
+  it("should pass a11y test", async () => {
+    const {container} = render(<Switch {...defaultSwitchProps} />);
+
+    await testA11y(container, {rules: {label: {enabled: false}}});
+  });
+
+  it("should run onToggle event handler correctly", () => {
+    render(<Switch {...defaultSwitchProps} />);
+
+    userEvent.click(screen.getByRole("switch"));
+
+    expect(defaultSwitchProps.onToggle).toHaveBeenCalledTimes(1);
+  });
+
+  it("should not run onToggle event handler when isDisabled is true", () => {
+    render(<Switch {...defaultSwitchProps} isDisabled={true} />);
+
+    jest.clearAllMocks();
+
+    userEvent.click(screen.getByRole("switch"));
+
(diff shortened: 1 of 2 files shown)


bilgeyucel (approved):

src/switch/switch.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    expect(defaultSwitchProps.onToggle).not.toHaveBeenCalled();
+  });
+
+  it("should add disabled attribute when isDisabled is true", () => {
+    render(<Switch {...defaultSwitchProps} isDisabled={true} />);
+
+    expect(screen.getByRole("switch")).toBeDisabled();
+  });
+  it("should add aria-readonly attribute when isDisabled is true", () => {
```
jamcry, line 49:
```suggestion
  });
  
  it("should add aria-readonly attribute when isDisabled is true", () => {

```

src/switch/switch.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+  };
+
+  it("should render correctly", () => {
+    render(<Switch {...defaultSwitchProps} />);
+  });
+
+  it("should pass a11y test", async () => {
+    const {container} = render(<Switch {...defaultSwitchProps} />);
+
+    await testA11y(container, {rules: {label: {enabled: false}}});
```
jamcry, line 23: Can you explain `{rules: {label: {enabled: false}}}`

src/switch/switch.test.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+    userEvent.click(screen.getByRole("switch"));
+
+    expect(defaultSwitchProps.onToggle).toHaveBeenCalledTimes(1);
+  });
+
+  it("should not run onToggle event handler when isDisabled is true", () => {
+    render(<Switch {...defaultSwitchProps} isDisabled={true} />);
+
+    jest.clearAllMocks();
```
jamcry, line 37: What does this line do?

Anlerkan: Since I used the onToggle mock function in the previous lines, i need to clear the mock calls and instances here for consistent testing. But, i think i can move this to jest config. So it will automatically clear mock calls and instances before every test.

src/switch/switch.test.tsx
```diff
@@ -0,0 +1,59 @@
...
+    render(<Switch {...defaultSwitchProps} isDisabled={true} />);
+
+    expect(screen.getByRole("switch")).toHaveAttribute("aria-readonly");
+  });
+
+  it("should add checked attribute when isToggledOn is true", () => {
+    render(<Switch {...defaultSwitchProps} />);
+
+    expect(screen.getByRole("switch")).toBeChecked();
+  });
```
mfarukkoc, line 58:
```suggestion
  });

  it("should not be checked when isToggledOn is false", () => {
    render(<Switch {...defaultSwitchProps} isToggledOn={false} />);

    expect(screen.getByRole("switch")).not.toBeChecked();
  });
```

Anlerkan: Thank you! I added tests for this case at https://github.com/Hipo/react-ui-toolkit/pull/136/commits/ffd6ee4e9d7df9398a54f51e48431ec2ff8b9c98

mfarukkoc (approved):

mucahit (approved):

PR #141, opened 2021-07-27
refactor(formField): Move children outside the label

### LINKED ISSUE #132: Typeahead: First tag is removed when clicked anywhere in box
https://user-images.githubusercontent.com/42540003/125934628-f107c4e1-fd41-4a8f-a45e-12a3d3ba188e.mp4

When the mouse enters the Typeahead box, it hovers the first tag, and the tag is removed on click.

mfarukkoc:
### Description
- Moved children prop outside the label element.

Fixes #132


jamcry:
I wanted to leave a comment for future reference:
This change was made at the first place because, if `TypeaheadSelect` is used inside a `FormField`, clicking the label causes auto-focus on the typeahead input, thus causing selected tag to be removed.

But we decided in the team meeting that moving `children` outside would be a breaking change, and we should try to find a better solution for the problem

PR #142, opened 2021-07-28
feat(list): Implement new list types

### LINKED ISSUE #137: `DescriptionList` component
In most of the projects, we create a custom `DescriptionList` component.
We better add this to toolkit as well.

- Maybe we can add a prop to the existing `List`, something like: 
  ```ts
  type: "unordered" | "ordered" | "description";
  ```
  then we can render the corresponding elements inside `List`.
- BUT, the shape of the item objects (`DescriptionTerm` and `ListItem`) are different. We should be aware of this.
- Here is an example `DescriptionTermShape`:
  ```ts
  export interface DescriptionTermShape {
    testid: string;
    title: string;
    description: React.ReactNode;
    descriptionFirst?: boolean;
    customClassName?: string;
  }
  ```

mfarukkoc:
### Description
- Implemented "description" and "ordered" list types.
- Create  `DescriptionTerm` component for the description list type.
- Add DescriptionTerm component to the `index.ts`

Closes #137


src/list/List.tsx
```diff
@@ -18,6 +18,7 @@ export interface ListProps<Item = any> {
     shouldDisplayEmptyState: boolean;
     emptyState: React.ReactNode;
   };
+  type?: "unordered" | "ordered" | "description";
 }
 
 function List<Item extends any>({
```
mucahit, line 24: This should also work (I haven't tested it)

```javascript
import "./_list.scss";

import React, {Fragment} from "react";
import classNames from "classnames";

export interface ListProps<Item = any> {
  items: Item[];
  children: (item: Item, testid: string, index?: number) => JSX.Element;
  listItemKeyGenerator?: (item: Item, testid: string) => string;
  testid?: string;
  role?: string;
  customClassName?: string;
  placeholderProps?: {
    shouldDisplayPlaceholder: boolean;
    placeholder: React.ReactNode;
  };
  emptyStateProps?: {
    shouldDisplayEmptyState: boolean;
    emptyState: React.ReactNode;
  };
  type?: "unordered" | "ordered" | "description";
}

function List<Item extends any>({
  items,
  children,
  customClassName,
  testid,
  role,
  listItemKeyGenerator,
  placeholderProps,
  emptyStateProps,
  type = "unordered"
}: ListProps<Item>) {
  const listClassName = classNames("list", customClassName);
  let ListTypeElement: keyof JSX.IntrinsicElements;

  if (type === "ordered") {
    ListTypeElement = "ol";
  } else if (type === "description") {
    ListTypeElement = "dl";
  } else {
    ListTypeElement = "ul";
  }

  return (
    <ListTypeElement className={listClassName} role={role}>
      {items.map((item: Item, index: number) => {
        const listItemTestId = `${testid}.item-${index}`;
        let key = listItemTestId;

        // @ts-ignore
        if (item && typeof item === "object" && item.id) {
          // @ts-ignore
          key = item.id;
        }

        if (listItemKeyGenerator) {
          key = listItemKeyGenerator(item, listItemTestId);
        }

        return <Fragment key={key}>{children(item, listItemTestId, index)}</Fragment>;
      })}

      {placeholderProps?.shouldDisplayPlaceholder && placeholderProps.placeholder}

      {!placeholderProps?.shouldDisplayPlaceholder &&
        emptyStateProps?.shouldDisplayEmptyState &&
        emptyStateProps.emptyState}
    </ListTypeElement>
  );
}

export default List;

```

mucahit (changes requested):

src/list/List.tsx
mfarukkoc: I tested and it is working, thanks.

src/list/description-term/DescriptionTerm.tsx
```diff
@@ -0,0 +1,52 @@
+import "./_description-term.scss";
+
+import React from "react";
+import classNames from "classnames";
+
+interface DescriptionTermProps {
+  testid?: string;
+  title: string;
+  description: React.ReactNode;
+  descriptionFirst?: boolean;
```
mucahit, line 10: This could be done with. What do you think?

```css
.description-term {
  display: flex;
  flex-direction: column-reverse;
}
```

or 

```css
.description-term {
  display: flex;
  flex-direction: column;
}

.description-term__title {
  order: 1;
}
```

mucahit (changes requested):

src/list/List.tsx
mucahit: Nice, thanks!

src/list/description-term/DescriptionTerm.tsx
mfarukkoc: We can do this but maybe usage with prop have other benefits. Let's ask @jamcry's opinion about this.

jamcry: I think this was defined as a prop because with this, in the same list, there may be some elements with the default order, some element with `descriptionFirst`. I think this would be harder if we made it with css, I believe we would have to pass classnames to each term.

```js
{
   id: "a",
   descriptionFirst: true
},
{
   id: "b",
   descriptionFirst: false
},
{
   id: "c",
   descriptionFirst: true
},
```

would be something like this:

```js
{
   id: "a",
      customClassName: "xx-list__reversed-order-term"
},
{
   id: "b",
},
{
  id: "c",
     customClassName: "xx-list__reversed-order-term"
 }
```

**But**, I don't think this has many use cases. We almost always use with all of them `descriptionFirst: true` when needed. Maybe we can discuss this next week. (I am off next Thursday, but I vote for CSS version as well 😄 )

mucahit (approved):

PR #146, opened 2021-08-10
refactor(toggle): Refactor toggle with input element

### LINKED ISSUE #144: Use `Checkbox` and `RadioInput` in `Toggle`
In `Toggle`, for better accessibility, we can use:
- Checkbox (if multi select)
- RadioInput (if single select)

mfarukkoc:
### Description
Refactored Toggle component with input element,

*  If the toggle is multi-select, use checkbox input
*  If the toggle is single select, use radio input 

> Note: I am unable to use `RadioInput` and `CheckboxInput` components because they require the `item` prop and using icons.

Closes #144

### DIFF
--- a/src/toggle/item/ToggleItem.tsx
+++ b/src/toggle/item/ToggleItem.tsx
@@ -28,12 +28,38 @@ function ToggleItem({
     "toggle-item--is-disabled": isDisabled
   });
 
+  if (canSelectMultiple) {
+    return (
+      <ListItem
+        testid={testid}
+        customClassName={toggleItemClassName}
+        clickableListItemProps={isDisabled ? undefined : {onClick: handleToggle}}>
+        <label className={"toggle-item__label"}>
+          <input
+            type={"checkbox"}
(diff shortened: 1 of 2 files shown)


src/toggle/item/ToggleItem.tsx
```diff
@@ -28,12 +28,38 @@ function ToggleItem({
...
-      {children}
+      <label className={"toggle-item__label"}>
+        <input
+          type={"radio"}
+          className={"toggle-radio-input"}
+          checked={isSelected}
+          disabled={isDisabled}
+        />
+        {children}
+      </label>
```
yasincaliskan, line 62: If we don't need more different changes let's just change the input type. Also, we should have only one `return` inside the components. Otherwise, it would be confusing.

```suggestion
      <label className={"toggle-item__label"}>
        <input
          type={canSelectMultiple ? "checkbox" : "radio"}
          className={"toggle-radio-input"}
          checked={isSelected}
          disabled={isDisabled}
        />
        {children}
      </label>
```

PR #157, opened 2021-10-08
build: I got changes in these files after I ran npm install with node 14.17 and npm 7.20

edizcelik:


yasincaliskan (commented):
I also faced a similar situation when I update the input test file. I guess it was because of run `prettify`. But I think it could be annoying for git history.

PR #169, opened 2021-10-28
test(toast): Write tests for toast component

edizcelik:


src/toast/toast.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+    userEvent.click(button);
+
+    expect(await screen.findByText(testToastContentText)).toBeVisible();
+
+    await waitFor(
+      () => {
+        expect(screen.queryByText(testToastContentText)).not.toBeInTheDocument();
+      },
+      {timeout: DEFAULT_TOAST_TIMEOUT}
```
jamcry, line 78: Is this working as expected? Can there be some kind of race condition, because we are closing the modal and checking if it was closed at the same time.

src/toast/toast.test.tsx
```diff
@@ -0,0 +1,219 @@
...
+    expect(await screen.findByText(testToastContentText)).toBeVisible();
+
+    const hideButton = await screen.findByText(hideToastButtonText);
+
+    userEvent.click(hideButton);
+
+    expect(screen.queryByText(testToastContentText)).not.toBeInTheDocument();
+  });
+
+  it("Displays toast with close button", async () => {
```
jamcry, line 189: do we have a convention about case messages? maybe we can say all of them should start with lowercase

jamcry (approved):
Looks great! Looks like we covered most of the possible cases 👍

PR #174, opened 2021-12-03
feat(tab): Add tests

yigiterdev:
- I'm not sure if test cases is enough or implemented correctly. I'm open for suggestions.

### DIFF
--- a/src/tab/tab.test.tsx
+++ b/src/tab/tab.test.tsx
@@ -1,5 +1,5 @@
 import React from "react";
-import {render, screen} from "@testing-library/react";
+import {fireEvent, render, screen} from "@testing-library/react";
 import "@testing-library/jest-dom";
 
 import {TabItem, TabProps} from "./Tab";
@@ -9,21 +9,21 @@ import {testA11y} from "../core/utils/test/testUtils";
 const tabItems: TabItem[] = [
   {
     id: "tab-item-1",
-    content: <div data-testid={"uncontrolled-tab.content-0"}>{"Tab Item 1"}</div>
+    content: <div data-testid={"uncontrolled-tab.content-0"}>{"Tab Item 0"}</div>
   },
   {
     id: "tab-item-2",
-    content: <div data-testid={"uncontrolled-tab.content-1"}>{"Tab Item 2"}</div>
+    content: <div data-testid={"uncontrolled-tab.content-1"}>{"Tab Item 1"}</div>
   },
   {
     id: "tab-item-3",
-    content: <div data-testid={"uncontrolled-tab.content-2"}>{"Tab Item 3"}</div>,
+    content: <div data-testid={"uncontrolled-tab.content-2"}>{"Tab Item 2"}</div>,
     isDisabled: true
   }
 ];
 
 // For uncontrolled tab component
-describe("<Tab/>", () => {
+describe("<UncontrolledTab/>", () => {
   const defaultUncontrolledTabProps: TabProps = {
     items: tabItems,
     children: [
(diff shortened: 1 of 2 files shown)


src/tab/tab.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+    expect(screen.getAllByRole("listitem").length).toEqual(tabItems.length);
+
+    for (let index = 0; index < tabItems.length; index++) {
+      const tabContent = screen.getByTestId(`uncontrolled-tab.content-${index}`);
+
+      expect(
+        screen.getByTestId(`uncontrolled-tab.header.item-${index}`)
+      ).toContainElement(tabContent);
+    }
```
edizcelik, line 65: Let's not check with the testid but instead we can check if the tab header with a certain text is added to the screen. This is to decouple the test from implementation details.

src/tab/tab.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+    render(<Tab {...defaultUncontrolledTabProps} />);
+  });
+
+  it("should pass a11y test", async () => {
+    const {container} = render(<Tab {...defaultUncontrolledTabProps} />);
+
+    await testA11y(container);
+  });
+
+  it("should render items correctly", () => {
```
edizcelik, line 54:
```suggestion
  it("should render header items correctly", () => {
```

src/tab/tab.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+    for (let index = 0; index < tabItems.length; index++) {
+      if (activeTabIndex === index) {
+        expect(screen.getByTestId(`uncontrolled-tab.body`)).toContain(children[index]);
+      }
+    }
+  });
+});
+
+// For controlled tab component
+describe("<Tab/>", () => {
```
edizcelik, line 94: I don't think there is a need to separate these out. But if we want to still keep them separate then we should give it a different name so that it isn't confusing.

src/tab/tab.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+    render(<Tab {...defaultUncontrolledTabProps} />);
+    const {activeTabIndex} = defaultUncontrolledTabProps;
+
+    for (let index = 0; index < tabItems.length; index++) {
+      if (activeTabIndex === index) {
+        expect(screen.getByTestId(`uncontrolled-tab.header.item-${index}`)).toHaveClass(
+          "tab-header-item--is-active"
+        );
+      }
+    }
```
edizcelik, line 78: Same here. Let's try to find a different way of selecting this element without depending on the implementation detail (if possible).

src/tab/tab.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+
+  it("should render active index correctly", () => {
+    render(<Tab {...defaultUncontrolledTabProps} />);
+    const {activeTabIndex, children} = defaultUncontrolledTabProps;
+
+    for (let index = 0; index < tabItems.length; index++) {
+      if (activeTabIndex === index) {
+        expect(screen.getByTestId(`uncontrolled-tab.body`)).toContain(children[index]);
+      }
+    }
```
edizcelik, line 89: Same here. We know the index and what should be displayed on the screen when that tab item was selected. So checking if that element is available on the screen will be a better test for this.

src/tab/tab.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+      </div>,
+      <div key={1} data-testid={"tab-views-1"}>
+        {"Following tab"}
+      </div>,
+      <div key={2} data-testid={"tab-views-2"}>
+        {"Disabled tab"}
+      </div>
+    ],
+    testid: "uncontrolled-tab",
+    initialActiveTabIndex: 1,
```
edizcelik, line 109: There should be a test that checks whether the correct active tab is rendered or not.

src/tab/tab.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+    onTabChange: jest.fn()
```
edizcelik, line 110: There should be a test that checks if this function is being called when a tab is trying to be changed

edizcelik (changes requested):
Thanks @yigiterdev! We should have the following tested as well:

- Render a tab component. Select a tab header item from the screen. Let's select item with id `tab-item-2`. Click on it. Check if the active classname was added to it. Then check if there is an element on the screen with a text `Tab Item 2`.

PR #178, opened 2022-01-13
feat(date-timer): Refactor `Countdown` as `DateTimer`

mucahit:
### Details
`Countdown` component has been repurposed as `DateTimer`, now it gets `range: Date[]` prop and countdowns to the greater date from the origin date. 

The other use case is Countup functionality, if you pass `timerType="up"`, `DateTimer` will expect a date greater than today so that it can count up to that date.

- `range` prop could be unordered, it will be handled by `calculateRemainingTimeBreakdown`
  - If there are more than 1 dates passed in the `range` prop, `calculateRemainingTimeBreakdown` will assign the greater date to `targetDate`
 

#### _This is a Breaking change_


src/date-timer/_date-timer.scss
```diff
@@ -1,11 +1,11 @@
-.countdown {
+.date-timer {
   display: grid;
   grid-template-columns: repeat(auto-fit, 60px);
 
   width: 100%;
 }
 
-.countdown-box {
+.date-timer-box {
```
edizcelik, line 8: Let's add the following property for `date-timer-box__count` class:

```
font-variant-numeric: tabular-nums;
```

It prevents the jump due to different digit widths

edizcelik (approved):

Issue #177, opened 2022-01-13
List: key uses array index by default


edizcelik:
We are currently using testid as the default key prop for each list item. This causes tracking issues if the provided items do not have an id property and if there was no key generator function was provided. I think we should initiate the key variable with undefined and then generate a unique string for it if there is no available `id` property in the item object or there was no `listItemKeyGenerator` provided as a prop.
https://github.com/Hipo/react-ui-toolkit/blob/main/src/list/List.tsx#L39

yasincaliskan:
Solution:
- Keep the initial key variable to undefined.
- Generate unique string with https://www.npmjs.com/package/uuid

PR #179, opened 2022-01-14
fix(list): Add uuid for each list item

yasincaliskan:
### Description
- Install `uuid` package to generate unique IDs.
- If not provided `listItemKeyGenerator` or `testid` props, generates uuid. 
- Also, generating a key with `testid` is not best practice. Because if list manipulate (add/remove an item or changing order) causes using wrong keys. 

### Related Issue: https://github.com/Hipo/react-ui-toolkit/issues/177

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,8 +100,10 @@
     "react-dom": "17.0.2"
   },
   "dependencies": {
+    "@types/uuid": "8.3.4",
     "classnames": "2.3.1",
-    "react-textarea-autosize": "8.3.2"
+    "react-textarea-autosize": "8.3.2",
+    "uuid": "8.3.2"
   },
   "peerDependencies": {
     "react": "17.0.2",
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "@types/react": "17.0.5",
     "@types/react-dom": "17.0.4",
     "@types/react-test-renderer": "17.0.1",
+    "@types/uuid": "8.3.4",
     "@typescript-eslint/eslint-plugin": "4.23.0",
     "@typescript-eslint/parser": "4.23.0",
     "axe-core": "4.2.0",
@@ -100,7 +101,6 @@
     "react-dom": "17.0.2"
   },
   "dependencies": {
-    "@types/uuid": "8.3.4",
     "classnames": "2.3.1",
     "react-textarea-autosize": "8.3.2",
     "uuid": "8.3.2"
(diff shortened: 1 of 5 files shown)


src/list/List.tsx
```diff
@@ -46,6 +47,8 @@ function List<Item extends any>({
 
         if (listItemKeyGenerator) {
           key = listItemKeyGenerator(item, listItemTestId);
```
jamcry, line 49: I think we should re-arrange the ifs:
```ts
   if (listItemKeyGenerator) {
          key = listItemKeyGenerator(item, listItemTestId);
          // @ts-ignore
    }  else if (item && typeof item === "object" && item.id) {
        // @ts-ignore
        key = item.id;
        }  else {
          key = uuidv4();
        }
```
Because I believe currently we don't use id in any case. We override the first `if` in 43 in the `else` in 50.
Also I think it's time to create a util for this and clean up the component function body:
```tsx
        return <Fragment key={generateListItemKey({
          item, listItemTestId, listItemKeyGenerator
          })}>
          {children(item, listItemTestId, index)}
        </Fragment>;
```

jamcry (changes requested):

src/list/List.tsx
yasincaliskan: Updated it, thank you!

jamcry (approved):

package.json
```diff
@@ -100,8 +100,10 @@
     "react-dom": "17.0.2"
   },
   "dependencies": {
+    "@types/uuid": "8.3.4",
```
mucahit, line 103: Should we move this to `devDependencies`?

src/list/util/listUtils.ts
```diff
@@ -0,0 +1,27 @@
...
+function generateListItemKey<Item = any>({
+  listItemKeyGenerator,
+  listItemTestId,
+  item
+}: {
+  listItemKeyGenerator: ListProps["listItemKeyGenerator"];
+  listItemTestId: string;
+  item: Item;
+}) {
+  let key = uuidv4();
```
mucahit, line 14: Can we move `uuidv4` to the `else` block of the below if? `uuidv4` should be our fallback solution, currently, it's creating id for every list item, then removes it (if any other id exists such as listItemKeyGenerator or item.id )

package.json
yasincaliskan: Yes, we should

src/list/util/listUtils.ts
yasincaliskan: Yes, it's a good point! Thank you 🙏🏻

mucahit (approved):
Looks good to me, thank you!

Anlerkan (approved):

PR #180, opened 2022-01-26
fix(numberInput):Cursor Positioning Fixed

tatata96:
**Description:**
When user tries to enter a new input by selecting the default value and typing the new input without deleting the current one, the cursor appears on the left side of the new  input when it should be on the right side, there is no issue when the default value is deleted and the new input is entered afterwards. You can test it in "Number Input with shouldFormatToLocaleString".

**Solution:**
A condition added to distinguish if one digit is edited, if so cursor positioning is applied, if not meaning that the input is a new one cursor position not modified.


src/form/input/Input.tsx
```diff
@@ -187,19 +187,28 @@ const Input = React.forwardRef<HTMLInputElement, InputProps>(
...
-              thousandthsSeparatorCount - 1
+            if (
+              String(value).length === finalEventValue.length + 1 ||
+              String(value).length === finalEventValue.length - 1
             ) {
-              caret += 1;
+              if (
+                prevValueThousandthsSeparatorCount ===
+                thousandthsSeparatorCount + 1
+                // eslint-disable-next-line no-magic-numbers
```
jamcry, line 197:
```suggestion
```
 Do we need this? It shouldn't give magic-numbers error for 1

jamcry (commented):
It looks good code-wise. But I'm not able to test it, so if you think you tested enough and it's not breaking anything else, I think it's good to go 👍

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@hipo/react-ui-toolkit",
-  "version": "1.0.0-alpha.5.1.0",
+  "version": "1.0.0-alpha.5.1.1",
```
edizcelik, line 3: Version change shouldn't have been done here. It is generally better if the person who is preparing the new published version to update the version of the package after all the changes are merged. Also version changes should be reflected on the package-lock file as well.

PR #188, opened 2022-02-22
fix(date-timer): Overall improvements on typing and functionality when both end of a range is provided

edizcelik:
- Changes titleMap type to be consistent with DateTimerItemID
- Moves `onEnd` callback handling to `useDateTimer` so that functionality can be shared with it as well if the users of the toolkit wants to implement their own custom timer UI from scratch.
- Regenerated package-lock with NPM 7. Updates the README as well.
- Changes the type of `DateTimerProps["range"]` to be `[Date, Date?]`
- Fixes the functionality when both ends of the range is provided.


mucahit (approved):
Thank you!

src/date-timer/util/dateTimerUtils.ts
```diff
@@ -26,7 +26,7 @@ function generateDateTimerItems(
     id: key,
     // eslint-disable-next-line no-magic-numbers
     count: dateTimerData[key].toString().padStart(2, "0"),
-    title: (titleMap ? titleMap[key.replace("s", "")] : undefined) || key
+    title: (titleMap ? titleMap[key] : undefined) || key
```
yasincaliskan, line 29: Can't it be like this?
```suggestion
    title: titleMap ? titleMap[key] : key
```

edizcelik: `titleMap[key]` may not exist as well

Anlerkan (approved):

yasincaliskan (approved):

PR #190, opened 2022-03-21
feat: Add Select, remove Dropdown

Anlerkan:
### Description
- `Select`, `SelectItem`, `SelectGroup`, `SelectContent`, `SelectTrigger`, `SelectItemList` implementation
- `useSelectClassName`, `useSelectKeyDown` hooks added
-  Used new `Select` instead of good old `Dropdown` on `TypeaheadSelect`  
- Added two new hooks to handle select logic: `useMultiSelect` and `useSingleSelect`

### Things to discuss
- [X]  How should we add `isMenuOpen` prop like we have before on `Dropdown` to `Select` ? 
- [x]  In this version, we have a prop to get selected option/options called `value`. It helps us to determine if the select is `multi-select` or `single-select` . We can handle it like this instead. Which one should we 

```tsx
 type SelectProps =
   | {
       children: React.ReactNode;
       role: SelectRole;
       options: Option[];
       selectedOption: Option | null;
       isMultiSelect: false;
       ...
     }
   | {
       children: React.ReactNode;
       role: SelectRole;
       options: Option[];
       selectedOptions: Option[];
       isMultiSelect: true;
       ...
     };
```

### TODO
- [X] We’ll provide `<SelectList/>` component
- [x] All of the Select components should have `ref`
- [X] We’ll test the options focus logic
- [x] `Options` prop doesn’t have to be a shape
- [X] Space press should select the focused option
- [X] isMenuOpen should be controlled or uncontrolled


edizcelik (changes requested):
A few things I noticed when checked the functionality over Storybook:
- It seems like we can highlight and focus on the disabled items with keyboard navigation and mouse over. I think the disabled items should be skipped on the keyboard navigation and should not be highlighted when hovered over.
- There seems to be no difference between the selected and hovered item states.
- I think the black border on the trigger can stay while the menu is open
- Not sure about the behavior of pressing TAB is correct while the menu is open. Currently hitting TAB when the menu is open changes the focused Item but after some item, the focus jumps to the next Select's trigger while the previous Select's menu is still kept open. I think when the TAB is hit, we should simply focus on the next focusable element and close the menu. We can also disable the TAB interaction when the menu is open and so they would first need to close the menu to focus on the next control element with TAB.  https://stackoverflow.com/questions/71355535/accessibility-question-do-select-dropdown-elements-need-to-be-tab-able-to-go-up 
- I wrote "tur" to the TypeaheadSelect's input and then when I try to focus on the input to change the keyword, it doesn't allow me. I think this is because we are changing the focused element to the first available option within the menu when we try to focus on the Select trigger. I think what we can do instead is to not focus on the first item when the menu is first opened, but only focus on the first item after menu is open and DOWN_ARROW key is hit.
- The first example's menu under Typeahead story is has z-index issue

src/select/content/_select-content.scss
```diff
@@ -0,0 +1,25 @@
...
+  border: 1px solid #dfe3ef;
+  border-radius: 6px;
+
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.05);
+
+  visibility: hidden;
+
+  &--is-visible {
+    visibility: visible;
+  }
```
yasincaliskan, line 24: How about using opacity with a transition instead? So that, we can change the transition on the project side.

```suggestion
  opacity: 0;
  
  transition: opacity 0.25s ease;

  &--is-visible {
    opacity: 1;
  }

```

src/select/item-list/SelectItemList.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import {Option} from "../util/selectTypes";
+import List, {ListElementType} from "../../list/List";
+import Select from "../Select";
+import {ListProps} from "../..";
+
+export type SelectItemListProps<T extends Option = Option> = {
+  options: T[];
+  customClassName?: string;
+  contentRenderer: (option: T) => React.ReactNode;
```
yasincaliskan, line 12: Shall we change the order? It is more readable imo.

```suggestion
  contentRenderer: (option: T) => React.ReactNode;
  customClassName?: string;
```

src/select/trigger/_select-trigger.scss
```diff
@@ -0,0 +1,29 @@
...
+
+  min-height: var(--select-trigger-height);
+
+  padding: 16px;
+
+  /* stylelint-disable-next-line color-no-hex */
+  border: 1px solid #dfe3ef;
+  border-radius: var(--select-trigger-border-radius);
+
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.05);
```
yasincaliskan, line 23: Not a big deal for sure, but we can remove it too

src/select/typeahead/_typeahead-select.scss
```diff
@@ -1,3 +1,7 @@
+.typeahead-select {
+  width: auto;
+}
```
yasincaliskan, line 3: AFAIK, the default width value is `auto`. So, we can remove it if we are not overriding something. Wdyt?

yasincaliskan (commented):
Looks awesome @Anlerkan , thanks! 🚀 
I just left really small comments for styling

jamcry:
I just tested the UX of the component, some things I noticed (we better fix these before merging the PR):

### Typeahead
- Selected option is removed from the list, but it can actually be focused with arrow keys and selected again and again

  https://user-images.githubusercontent.com/22727931/188152167-9b2b4155-e2cc-4712-b2f5-db2f4e804e8a.mov

  - Reproduction (maybe the video below is more clear):
    1. Select an item, say "English". it's hidden from the list now
    2. Navigate between remaining items with up and down arrow keys. When you come to the place where the selected option was, the focus is on that item, and if you hit Enter, it's selected  
- Typeahead menu won't be closed when I click on the header part (This is same for normal Select as well. If menu is open and you click on header button, it's not closed)

  https://user-images.githubusercontent.com/22727931/188152592-98ede789-429e-4256-bc00-34c22f2f39d6.mov
- Even after selecting multiple number of items possible, menu is opened

  https://user-images.githubusercontent.com/22727931/188152789-0b236f11-281b-4793-bfd0-1cd0fb168a1c.mov

- The style of the tags (selected item tags in the header box) are a bit weird and they don't look vertically centered inside the box (can be seen in the previous videos)

### Select
- We should improve TAB behavior as we discussed on the meeting
- (Video 1) I believe if the focus on the select header button, or on the menu, if I type some letters, it should search it on the list (We had this feature on the prev. Dropdown component, and native select also has this) 

  https://user-images.githubusercontent.com/22727931/188153459-6b9dd59e-0952-4538-99dc-c55368e9f433.mov
- (Image 1) I think we should add some kind of selected item marker for multi-select. Now, it's a bit confusing. Both selected items and focused items look the same
  Screenshot 2022-09-02 at 15 13 59

PR #197, opened 2022-06-23
refactor(toast): Fix unnecessarily re-rendering all children of toast context on every state change

jamcry:
### What is the problem?
  -  In our apps, we mostly only care Toast state `dispatch` function. But in the current implementation, since `useToaster` returns both state and dispatch in the same object, dispatch function will "change" on every state change, and this will cause re-render in the components that use `dispatch`.

### Possible (implemented) Solution
  - Separating state and dispatch contexts, so that state changes won't trigger a change in `dispatch`
 
### Result (before vs after) [green border means that element is re-rendered]
I tried on my local with `npm-link`, here is the result:
https://user-images.githubusercontent.com/22727931/175345592-35eadfd1-6b05-4a3d-9334-7320a0e95f32.mov
Only necessary elements re-renders. As expected.

**If the app is only using `useToaster`, it won't be a breaking change and won't require any changes on the project**

### DIFF
--- a/src/toast/ToastProvider.tsx
+++ b/src/toast/ToastProvider.tsx
@@ -6,8 +6,8 @@ import toastReducer from "./util/toastReducer";
 import {ToastAction, ToastContextState} from "./util/toastTypes";
 import {isNonNegativeNumber} from "../core/utils/number/numberUtils";
 
-const ToastStateContext = createContext<ToastContextState>(initialToastState);
-const ToastDispatchContext = createContext<React.Dispatch<ToastAction>>(() => undefined);
+const ToastStateContext = createContext<null | ToastContextState>(null);
+const ToastDispatchContext = createContext<null | React.Dispatch<ToastAction>>(null);
 
 ToastDispatchContext.displayName = "ToastDispatchContext";
 ToastStateContext.displayName = "ToastStateContext";
--- a/src/toast/ToastProvider.tsx
+++ b/src/toast/ToastProvider.tsx
@@ -6,12 +6,11 @@ import toastReducer from "./util/toastReducer";
 import {ToastAction, ToastContextState} from "./util/toastTypes";
 import {isNonNegativeNumber} from "../core/utils/number/numberUtils";
 
-const ToastContext = createContext<[ToastContextState, React.Dispatch<ToastAction>]>([
-  initialToastState,
-  () => undefined
-]);
+const ToastStateContext = createContext<ToastContextState>(initialToastState);
+const ToastDispatchContext = createContext<React.Dispatch<ToastAction>>(() => undefined);
 
-ToastContext.displayName = "ToastContext";
+ToastDispatchContext.displayName = "ToastDispatchContext";
+ToastStateContext.displayName = "ToastStateContext";
 
 interface ToastContextProviderProps {
   children: React.ReactNode;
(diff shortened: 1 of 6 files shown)


src/toast/util/toastHooks.ts
```diff
@@ -2,27 +2,24 @@ import {useCallback, useContext} from "react";
...
+ * @returns {Object} Current value of ToastContextState
  */
-function useToastContext() {
-  const context = useContext(ToastContext);
+function useToastState(): ToastContextState {
+  const state = useContext(ToastStateContext);
 
-  if (!context) {
+  if (!state) {
     throw new Error("Trying to consume ToastContext outside of its provider.");
```
edizcelik:
```suggestion
    throw new Error("Trying to consume ToastStateContext outside of its provider.");
```

src/toast/util/toastHooks.ts
```diff
@@ -2,27 +2,24 @@ import {useCallback, useContext} from "react";
...
-  return context;
+  return state;
 }
 
-/**
- * @returns {function} ToastContext's state reducer's dispatch function
- */
 function useToaster() {
-  const dispatch = useToastContext()[1];
+  const dispatch = useContext(ToastDispatchContext);
```
edizcelik, line 23:
```suggestion
  const dispatch = useContext(ToastDispatchContext);

  if (!dispatch) {
     throw new Error("Trying to consume ToastDispatchContext outside of its provider");
  }
```

src/toast/ToastProvider.tsx
```diff
@@ -6,12 +6,11 @@ import toastReducer from "./util/toastReducer";
 import {ToastAction, ToastContextState} from "./util/toastTypes";
 import {isNonNegativeNumber} from "../core/utils/number/numberUtils";
 
-const ToastContext = createContext<[ToastContextState, React.Dispatch<ToastAction>]>([
-  initialToastState,
-  () => undefined
-]);
+const ToastStateContext = createContext<ToastContextState>(initialToastState);
+const ToastDispatchContext = createContext<React.Dispatch<ToastAction>>(() => undefined);
```
edizcelik, line 10: Its better to initialize the Context values with `null` or `undefined` so that the Provider is properly set up and the consumers work as expected to avoid unnecessary bugs. As there are custom hooks that we provide for consuming the context value, the nullish cases will be handled there and raise an error if the context is not being used properly.

```suggestion
const ToastDispatchContext = createContext<null | React.Dispatch<ToastAction>>(null);
```

Please update the ToastStateContext as well so that it is initialized with `null`

edizcelik (changes requested):
This approach is the way to go when using Contexts 🥇 I had a few comments about how we can make the consumers safer

edizcelik:
By the way, we can update the file structure here. We recently started using this convention on Tinyman and it looks quite nice:

```
.../ToastContext.tsx     <- this includes Contexts and the ToastContextProvider
.../ToastContext.utils.ts      <-  this includes the custom hooks for the consumers
.../ToastContext.reducer.ts      <- includes the reducer and the state stuff
.../ToastContext.constants.ts    <- includes constants related to the Context
.../ToastContext.types.ts      <- includes types related to the Context (if any)
```

jamcry:
> By the way, we can update the file structure here. We recently started using this convention on Tinyman and it looks quite nice:
> 
> ```
> .../ToastContext.tsx     <- this includes Contexts and the ToastContextProvider
> .../ToastContext.utils.ts      <-  this includes the custom hooks for the consumers
> .../ToastContext.reducer.ts      <- includes the reducer and the state stuff
> .../ToastContext.constants.ts    <- includes constants related to the Context
> .../ToastContext.types.ts      <- includes types related to the Context (if any)
> ```

I think it's a good idea to make these files more organized 👍 
It will cause changes at a lot of files because toast has three contexts now: `ToastDispatchContext`, `ToastStateContext` and `ToastItemContext`.

I'll create an issue, we can do this refactor in another PR, also for `ToggleContext` to make all consistent

edizcelik (approved):

akgunberk (approved):

Issue #206, opened 2022-11-03
Select: Make `Option` type generic


jamcry:
Previously, `DropdownOption` was generic, so we were able to type its `Id` and `Context`, like this: `DropdownOption<UserType>`. But with Select changes, we lost this feature.

Before:
```tsx
export interface DropdownOption<Id = string, Context = any> {
  id: Id;
  title: string;
  customClassName?: string;
  CustomContent?: JSX.Element;
  icon?: React.ReactNode;
  subtitle?: string;
  context?: Context;
  isDisabled?: boolean;
}
```

Now:
```tsx
interface Option {
  id: string;
  isDisabled?: boolean;
}
```

We should try to restore this.

PR #214, opened 2022-11-28
fix(select): Create a generic Id type for Option interface

gulcinuras:
- Remove `title` from `TypeaheadSelectOption` and use `contentRenderer` prop as children to `Select.Item` in `TypeaheadSelect`
- Create an enum for Language options and use it in `Select` and `TypeaheadSelect` components
- Use `contentRenderer` for `tag` content as we removed the `title` prop
- Remove `initialKeyword` and use `controlledKeyword` to set initial and changeable value
- Set `canSelectMultiple` to false if `options.length` is not bigger than 1
- Add `testid` to use in `typeahead-select.test`
- Add `onClick` optional prop to `TypeheadSelectTrigger` to set the correct `classname` when the menu is open
- `setMenuVisibility` to false if `shouldCloseOnSelect` is true, so that the `classnames` are correct
- Remove input styles when `TypeaheadSelectInput` is used inside of the trigger
- Add `testid` to `TypeheadSelectTrigger` to test it easily
- Fix most of the test cases for `typeahead-select`
         - Fix classnames
         - I used `screen.findByText` function to find the option but I can update it

Note:
- In my opinion, having two similar classnames for `TypeaheadSelect` and Select is a bit unnecessary. We can just add typeahead-select as main classname and override select classnames if necessary.
- I couldn't fix `a11y` test case for the `typeahead-select`.
- As indicated in #206 issue I added `Id` as generic type


src/select/util/selectTypes.ts
```diff
@@ -1,13 +1,19 @@
...
-  title: string;
-}
+// TypeaheadSelectOption is intentionally empty. It happens not
+// to have more properties than Option, but this may
+// change in the future, and it helps to have a TypeaheadSelectOption
+// interface that people can use. Therefore the no-empty-interface is disabled
+// rule for this declaration:
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+interface TypeaheadSelectOption extends Option {}
```
Anlerkan, line 16: We can use type instead of an interface to avoid disabling this rule.

```suggestion
type TypeaheadSelectOption = Option;
```

src/select/util/selectTypes.ts
```diff
@@ -1,13 +1,19 @@
 import React from "react";
 
-interface Option {
-  id: string;
+interface Option<Id = string> {
+  id: Id;
+  title: React.ReactNode;
```
Anlerkan, line 5: Why did we add `title` property here?

src/select/util/selectTypes.ts
gulcinuras: I tried the way you said, but it didn't let me set a generic type to `TypeaheadSelectOption` 🤔

src/select/util/selectTypes.ts
gulcinuras: I think it can be useful for some cases, but I can set it as optional or remove it completely.

src/select/util/selectTypes.ts
```diff
@@ -1,13 +1,14 @@
...
   isDisabled?: boolean;
 }
 
-interface TypeaheadSelectOption extends Option {
-  title: string;
-}
+type TypeaheadSelectOption<Id = string> = Omit<Option, "id"> & {
+  id: Id;
+};
```
Anlerkan, line 11: I guess that would also work
```suggestion
type TypeaheadSelectOption<Id = string> = Option<Id>;
```

ErenYLCN (dismissed):

edizcelik (approved):

tatata96 (approved):

src/select/typeahead/TypeaheadSelect.tsx
Anlerkan, line 1: I think we should have made a refactor for that component. It looks too complex :(

src/select/typeahead/trigger/TypeheadSelectTrigger.tsx
```diff
@@ -13,15 +13,20 @@ export interface TypeheadSelectTriggerProps {
...
   customClassName,
-  input
+  input,
+  onClick
 }: TypeheadSelectTriggerProps) {
   return (
-    <Select.Trigger customClassName={"typeahead-select-trigger"}>
+    <Select.Trigger
+      customClassName={"typeahead-select-trigger"}
+      testid={"TypeaheadSelectTrigger"}
```
Anlerkan, line 28: Why do we need this testid?

Issue #209, opened 2022-11-10
NumberInput: It should provide `number-input` as a customClassName to Input


edizcelik:
This can allow us to customize NumberInput styling more easily on projects

PR #217, opened 2022-12-08
feat(input/number): Add className `number-input`

ErenYLCN:
- Related issue #209

### DIFF
--- a/src/form/input/number/NumberInput.tsx
+++ b/src/form/input/number/NumberInput.tsx
@@ -50,7 +50,14 @@ const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>((props,
   });
 
   return (
-    <Input ref={ref} type={"text"} onChange={handleChange} value={finalValue} {...rest} />
+    <Input
+      ref={ref}
+      customClassName={"number-input"}
+      type={"text"}
+      onChange={handleChange}
+      value={finalValue}
+      {...rest}
+    />
   );
 
   function handleChange(event: React.SyntheticEvent<HTMLInputElement>) {


gulcinuras (dismissed):

tatata96 (dismissed):

src/form/input/number/NumberInput.tsx
```diff
@@ -50,7 +50,14 @@ const NumberInput = React.forwardRef<HTMLInputElement, NumberInputProps>((props,
...
   return (
-    <Input ref={ref} type={"text"} onChange={handleChange} value={finalValue} {...rest} />
+    <Input
+      ref={ref}
+      customClassName={"number-input"}
+      type={"text"}
+      onChange={handleChange}
+      value={finalValue}
+      {...rest}
+    />
```
Anlerkan, line 60: I'm afraid it won't be enough.
- If I provide `customClassName` to `NumberInput`, this class name will be overridden because `...rest` has `customClassName`. Although we don't have any specific style for the number input, we should keep that class name.
- I think the `NumberInput` should be able to take custom class names, so we should have both.

Here is my suggestion:

```suggestion
    <Input
      ref={ref}
      customClassName={classNames("number-input", customClassName)}
      type={"text"}
      onChange={handleChange}
      value={finalValue}
      {...rest}
    />
```

Anlerkan (changes requested):

Anlerkan (approved):

edizcelik (approved):

yasincaliskan (approved):

PR #223, opened 2023-01-18
fix(tab):generic id type added to TabItem

tatata96:
**Related Issue;**
- #172

### DIFF
--- a/src/tab/Tab.tsx
+++ b/src/tab/Tab.tsx
@@ -6,15 +6,15 @@ import classNames from "classnames";
 import TabHeaderItem from "./header/item/TabHeaderItem";
 import List from "../list/List";
 
-export type TabItem = {
-  id: string;
+export type TabItem<ID = string | number> = {
+  id: ID;
   content: React.ReactNode;
   icon?: React.ReactNode;
   isDisabled?: boolean;
 };
 
-interface UncontrolledTabProps {
-  items: TabItem[];
+interface UncontrolledTabProps<ID> {
+  items: TabItem<ID>[];
   children: React.ReactNode[];
   testid?: string;
   initialActiveTabIndex?: number;
@@ -26,16 +26,16 @@ interface UncontrolledTabProps {
 // and initialActiveTabIndex should be undefined
 type ControlledTabProps =
   | {
-      activeTabIndex: number;
-      onTabChange: (index: number) => void;
-      initialActiveTabIndex?: number;
-    }
+    activeTabIndex: number;
+    onTabChange: (index: number) => void;
+    initialActiveTabIndex?: number;
+  }
   | {
-      activeTabIndex?: number;
-      onTabChange?: (index: number) => void;
-    };
+    activeTabIndex?: number;
+    onTabChange?: (index: number) => void;
+  };
 
-export type TabProps = ControlledTabProps & UncontrolledTabProps;
+export type TabProps = ControlledTabProps & UncontrolledTabProps<string | number>;
 
 function Tab({
   testid,
@@ -70,9 +70,9 @@ function Tab({
       <div className={"tab__body"} data-testid={`${testid}.body`}>
         {
           children[
-            activeTabIndexFromProps === undefined
-              ? activeTabIndex
-              : activeTabIndexFromProps
+          activeTabIndexFromProps === undefined
+            ? activeTabIndex
+            : activeTabIndexFromProps
           ]
         }
       </div>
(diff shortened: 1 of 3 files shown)


gulcinuras (commented):
@tatata96 I asked a small change, but we can also discuss if that makes sense or not 😸 
Also, I noticed that in `12-Tab.stories` we are not providing the `initialState` in a `Record` type, so that it gives a type error.

Therefore, I think we can update the `initialState` on line 39 like this:
```
<StateProvider initialState={{index: 0}}>
```
and update the other places we use that state.

src/tab/Tab.tsx
```diff
@@ -6,15 +6,15 @@ import classNames from "classnames";
...
+  id: ID;
   content: React.ReactNode;
   icon?: React.ReactNode;
   isDisabled?: boolean;
 };
 
-interface UncontrolledTabProps {
-  items: TabItem[];
+interface UncontrolledTabProps<ID> {
+  items: TabItem<ID>[];
```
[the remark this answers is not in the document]
ErenYLCN: I think we can discuss this with the team

ErenYLCN (approved):

src/tab/Tab.tsx
```diff
@@ -6,15 +6,15 @@ import classNames from "classnames";
 import TabHeaderItem from "./header/item/TabHeaderItem";
 import List from "../list/List";
 
-export type TabItem = {
-  id: string;
+export type TabItem<ID = string | number> = {
```
Anlerkan, line 9: In this version, we can pass boolean or false to the generic type. We can even pass array :D Should we? IMO, we should be able to pass string or number only.
```suggestion
export type TabItem<ID extends string | number = string | number> = {
```

tatata96: You're right. I missed it and gave default types only, I should have limited it using constraints. 
Thank you for noticing ✌️

gulcinuras (approved):

edizcelik (approved):

src/tab/Tab.tsx
```diff
@@ -6,15 +6,15 @@ import classNames from "classnames";
 import TabHeaderItem from "./header/item/TabHeaderItem";
 import List from "../list/List";
 
-export type TabItem = {
-  id: string;
+export type TabItem<ID extends string | number = string | number> = {
```
yasincaliskan, line 9: Just wonder, if the ID can be a string or number, should we extend it with these types? Wouldn't be okay like this? 🤔 
```suggestion
export type TabItem<ID = string | number> = {
```

yasincaliskan (commented):
🎉

src/tab/Tab.tsx
tatata96: With the current version the type of ID can only string or number (same as your suggestion), but if you don't specify a type argument, the default type will be one of those two.

stories/12-Tab.stories.tsx
```diff
@@ -45,21 +45,21 @@ storiesOf("Tab", module).add("Tab", () => (
...
           </div>
 
           <Tab
             items={tabItems}
-            activeTabIndex={state}
+            activeTabIndex={state.index}
             onTabChange={(index) => {
               console.log("tab changed to index: ", index);
-              setState(index);
+              setState({index: index});
```
Anlerkan, line 62: I think we can use shorthand version
```suggestion
              setState({index});
```

src/tab/header/item/TabHeaderItem.tsx
```diff
@@ -7,7 +7,7 @@ import {TabItem} from "../../Tab";
 import ListItem from "../../../list/item/ListItem";
 
 type TabHeaderItemProps = {
-  tab: TabItem;
+  tab: TabItem<string | number>;
```
Anlerkan, line 10: Shouldn't we get the ID shape from `TabHeaderItemProps` ?

src/tab/Tab.tsx
```diff
@@ -26,16 +26,16 @@ interface UncontrolledTabProps {
...
   | {
-      activeTabIndex?: number;
-      onTabChange?: (index: number) => void;
-    };
+    activeTabIndex?: number;
+    onTabChange?: (index: number) => void;
+  };
 
-export type TabProps = ControlledTabProps & UncontrolledTabProps;
+export type TabProps = ControlledTabProps & UncontrolledTabProps<string | number>;
```
Anlerkan, line 38: Also here, should not we derive the id shape from `TabProps`?

Anlerkan (changes requested):

yasincaliskan (approved):

jamcry (approved):

PR #224, opened 2023-01-18
fix(tab):id attribute added to tab wrapper

tatata96:
**Related Issue;**
- #213

### DIFF
--- a/src/tab/Tab.tsx
+++ b/src/tab/Tab.tsx
@@ -27,14 +27,14 @@ interface UncontrolledTabProps {
 // and initialActiveTabIndex should be undefined
 type ControlledTabProps =
   | {
-    activeTabIndex: number;
-    onTabChange: (index: number) => void;
-    initialActiveTabIndex?: number;
-  }
+      activeTabIndex: number;
+      onTabChange: (index: number) => void;
+      initialActiveTabIndex?: number;
+    }
   | {
-    activeTabIndex?: number;
-    onTabChange?: (index: number) => void;
-  };
+      activeTabIndex?: number;
+      onTabChange?: (index: number) => void;
+    };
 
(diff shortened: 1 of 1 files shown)


src/tab/Tab.tsx
```diff
@@ -70,9 +72,9 @@ function Tab({
       <div className={"tab__body"} data-testid={`${testid}.body`}>
         {
           children[
-            activeTabIndexFromProps === undefined
-              ? activeTabIndex
-              : activeTabIndexFromProps
+          activeTabIndexFromProps === undefined
+            ? activeTabIndex
+            : activeTabIndexFromProps
```
jamcry, line 77: The indentation doesn't look correct,
content of `[]` should be deeper, why was it changed in the first place? 🤔 I believe we already have eslint rules for the toolkit, it should take care of this.

Same for the diff in `ControlledTabProps`

src/tab/Tab.tsx
```diff
@@ -19,21 +19,22 @@ interface UncontrolledTabProps {
...
+    onTabChange: (index: number) => void;
+    initialActiveTabIndex?: number;
+  }
   | {
-      activeTabIndex?: number;
-      onTabChange?: (index: number) => void;
-    };
+    activeTabIndex?: number;
+    onTabChange?: (index: number) => void;
+  };
```
jamcry, line 37: This is totally unrelated to the issue, but I was trying to understand this comment and the type and I couldn't.
Do you (or anyone) know what is our purpose with this type? Why do we need the second part?
```tsx
    activeTabIndex?: number;
    onTabChange?: (index: number) => void;
```

(BTW, code formatting here also looks weird as I mentioned in the other comment)

edizcelik (approved):

yasincaliskan (approved):

Issue #171, opened 2021-11-18
Tab: Create a separate component for the header part


edizcelik:
I am interested only using the header of the tab component. I believe we can separate this part into a component on its own so that it can be used to compose a more flexibility tab structures. For example, in Tinyman, we are only interested in the functionality of the header and use it like a filter on a table view, so the tab body will be the same for each item on the tab.

PR #225, opened 2023-01-20
fix(tab):Separate tab body

tatata96:
**Related Issue;**
- #171

### DIFF
--- a/src/tab/Tab.tsx
+++ b/src/tab/Tab.tsx
@@ -19,6 +19,7 @@ interface UncontrolledTabProps {
   testid?: string;
   initialActiveTabIndex?: number;
   customClassName?: string;
+  shouldDisplayTabBody?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


gulcinuras (commented):
@tatata96 as far as I understand from [the mentioned issue](https://github.com/Hipo/react-ui-toolkit/issues/171), we should have a compound components (could be named as `TabHeader` and `TabBody`) which would allow us to customize easier. So, I think we should follow a similar atomized approach as we have in the `Select` component. What do you think?

ErenYLCN (changes requested):
I agree with @gulcinuras here, I think we should at least have a `TabHeader` component to completely fulfill the issue

yasincaliskan:
I also agree with @gulcinuras. We should have `TabHeader` and `TabBody` compound components. We can also check `Toggle` component, it is simpler than `Select`.

Issue #193, opened 2022-04-06
Tab: Select the first item if activeTabIndex is out of bounds


edizcelik:
After a Tab component is rendered on the screen, we can modify the `items` prop and pass in a new array. For example, we might display one tab on larger screens while two tabs on the smaller screens. If the second item was selected as active and then the screen size is changed so that the `items` array has only one element in it, `activeTabIndex` becomes out of bounds and we don't display anything as the tab body. In this case, I think we can simply fallback and set the activeTabIndex as 0.

PR #226, opened 2023-01-27
fix(tab): out of bounds issue for screen size change 

tatata96:
**Related Issue;**
- #193 

I tried recreate problem by updating the `items` (it had 5 elements, I changed it to 2) for screens smaller than 400, using `useWindowSize` (this will be removed, only there for testing purposes). 

https://user-images.githubusercontent.com/53620159/215018788-d145f0d4-380b-4ad1-97a8-65130e5b4ce0.mov

After adding the out of bounds check and setting `index` to 0:

https://user-images.githubusercontent.com/53620159/215018981-7921809a-4271-4f37-8599-da6b75a606af.mov

### DIFF
--- a/src/core/utils/hooks/useWindowSize.tsx
+++ b/src/core/utils/hooks/useWindowSize.tsx
@@ -5,17 +5,16 @@ interface Size {
   height: number | undefined;
 }
 function useWindowSize(): Size {
-
   const [windowSize, setWindowSize] = useState<Size>({
     width: undefined,
-    height: undefined,
+    height: undefined
   });
 
   useEffect(() => {
     function handleResize() {
       setWindowSize({
         width: window.innerWidth,
-        height: window.innerHeight,
+        height: window.innerHeight
       });
     }
     window.addEventListener("resize", handleResize);
--- a/src/core/utils/hooks/useWindowSize.tsx
+++ b/src/core/utils/hooks/useWindowSize.tsx
@@ -1,0 +1,28 @@
+import {useState, useEffect} from "react";
+// Define general type for useWindowSize hook, which includes width and height
+interface Size {
+  width: number | undefined;
+  height: number | undefined;
+}
+function useWindowSize(): Size {
+
+  const [windowSize, setWindowSize] = useState<Size>({
+    width: undefined,
+    height: undefined,
+  });
+
+  useEffect(() => {
+    function handleResize() {
+      setWindowSize({
+        width: window.innerWidth,
+        height: window.innerHeight,
+      });
+    }
+    window.addEventListener("resize", handleResize);
+    handleResize();
+    return () => window.removeEventListener("resize", handleResize);
+  }, []);
+  return windowSize;
+}
+
+export default useWindowSize;
--- a/src/tab/Tab.tsx
+++ b/src/tab/Tab.tsx
@@ -49,9 +49,8 @@ function Tab({
   const [activeTabIndex, setActiveTabIndex] = useState(initialActiveTabIndex);
   const tabClassName = classNames("tab", customClassName);
 
-  let activeTabIndexToUse =
-    activeTabIndexFromProps === undefined ? activeTabIndex : activeTabIndexFromProps;
-
+  let activeTabIndexToUse = activeTabIndexFromProps || activeTabIndex;
+  
   /**
    * In case activeTabIndex is out of bounds, we need to reset it to 0.
    * This can happen if items are changes after the mount
--- a/src/tab/Tab.tsx
+++ b/src/tab/Tab.tsx
@@ -26,14 +26,14 @@ interface UncontrolledTabProps {
 // and initialActiveTabIndex should be undefined
 type ControlledTabProps =
   | {
-    activeTabIndex: number;
-    onTabChange: (index: number) => void;
-    initialActiveTabIndex?: number;
-  }
+      activeTabIndex: number;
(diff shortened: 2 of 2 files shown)


jamcry (commented):
Hi @tatata96 , I realized code formatting on your PRs are a bit weird. Can you make sure your VSCode/IDE is using the correct linter/formatter?

For example, when I save this file, I get this diff (it formats according to our linter config files):
Screenshot 2023-01-27 at 12 47 36

Your ide should be using the local prettier config
Screenshot 2023-01-27 at 12 49 01

jamcry:
@tatata96 I believe issue link should be this: https://github.com/Hipo/react-ui-toolkit/issues/193

src/tab/Tab.tsx
```diff
@@ -49,6 +49,13 @@ function Tab({
   const [activeTabIndex, setActiveTabIndex] = useState(initialActiveTabIndex);
   const tabClassName = classNames("tab", customClassName);
 
+  let activeTabIndexToUse = activeTabIndexFromProps === undefined
+    ? activeTabIndex
+    : activeTabIndexFromProps;
+
+  // check if active index is out of bounds
+  activeTabIndexToUse = activeTabIndexToUse > items.length ? 0 : activeTabIndexToUse;
```
jamcry, line 57:
```suggestion
  const [activeTabIndex, setActiveTabIndex] = useState(initialActiveTabIndex);

  // If activeTabIndex is uncontrolled, we need to use the local state
  let activeTabIndexToUse =
    activeTabIndexFromProps === undefined ? activeTabIndex : activeTabIndexFromProps;

  /**
   * In case activeTabIndex is out of bounds, we need to reset it to 0.
   * This can happen if items are changes after the mount
   */
  if (activeTabIndexToUse > items.length) {
    activeTabIndexToUse = 0;
  }

  const tabClassName = classNames("tab", customClassName);
```

What do you think about this version?
- I think it's better if we explain the things here better, it can be confusing since we have 3 variables for the same value (activeTabIndex, activeTabIndexFromProps, activeTabIndexToUse)
- `a = condition ? 0 : a` actually includes a single case (set 0 if `condition`), so I believe no need for ternary(?) here and an if block can make it more readable

src/core/utils/hooks/useWindowSize.tsx
```diff
@@ -0,0 +1,28 @@
+import {useState, useEffect} from "react";
```
jamcry, line 1: > using useWindowSize (this will be removed, only there for testing purposes).

I saw this part, but this hook is not actually used at the moment, so maybe you can remove from remote, but keep it on your local if you need it

We could add an example of changing `items` to the story, but I don't think that's necessary

tatata96: I added that for testing purposes, it will be removed before the pr is merged

src/tab/Tab.tsx
```diff
@@ -49,6 +49,13 @@ function Tab({
   const [activeTabIndex, setActiveTabIndex] = useState(initialActiveTabIndex);
   const tabClassName = classNames("tab", customClassName);
 
+  let activeTabIndexToUse = activeTabIndexFromProps === undefined
+    ? activeTabIndex
+    : activeTabIndexFromProps;
```
Anlerkan, line 54: Would that work?

```suggestion
  let activeTabIndexToUse = activeTabIndexFromProps || activeTabIndex
```

Anlerkan (commented):
Not related to that issue but it came to mind while thinking about it,
Maybe we can simplify the `handleChangeActiveTab` function.
Here is my suggestion:

`const tabChangeHandler = onTabChange || setActiveTabIndex;`

```tsx
  function handleChangeActiveTab(index: number) {
    if (index !== activeTabIndexToUse) {
      tabChangeHandler(index);
    }
  }
```
What do you think? @tatata96 @Hipo/frontend-team

src/tab/Tab.tsx
tatata96: I agree, it's way more readable. Thanks 🙏

src/tab/Tab.tsx
tatata96: Yes, this works 👍

tatata96:
> 

I think yours is more readable but it may cause issues for cases where we want to use `onTabChange` for uncontrolled tabs.
I think we can use your version of `handleChangeActiveTab` but change `tabChangeHandler` like below maybe: 

`const tabChangeHandler = activeTabIndexFromProps && onTabChange ? onTabChange : setActiveTabIndex;`

What do you think? @Anlerkan

src/tab/Tab.tsx
edizcelik: What if `activeTabIndexFromProps` is 0? Does it still work as expected?

src/tab/Tab.tsx
```diff
@@ -49,6 +49,16 @@ function Tab({
...
 
+  let activeTabIndexToUse = activeTabIndexFromProps || activeTabIndex;
+  
+  /**
+   * In case activeTabIndex is out of bounds, we need to reset it to 0.
+   * This can happen if items are changes after the mount
+   */
+  if (activeTabIndexToUse > items.length) {
+    activeTabIndexToUse = 0;
+  }
```
edizcelik, line 60: This should be reflected to the state and also we should call `onTabChange` if the Tab is being used as a controlled component

edizcelik (changes requested):

PR #227, opened 2023-02-20
fix(select/options): Set options to the state with dispatch if the length is not equal

gulcinuras:
- After setting `props.options` as initial `options` value in `selectOwnState`, async options are not added in the options array. So the async SelectItem has an index value of -1 which is equal to the focusOptionIndex. That was why async items were highlighted.


edizcelik:
I guess only the last two commits belongs to this PR

src/select/util/selectTypes.ts
```diff
@@ -50,7 +49,8 @@ type SelectValue<T extends Option> = T | T[] | null;
 
 type SelectStateAction =
   | {type: "TOGGLE_MENU_VISIBILITY"}
-  | {type: "SET_FOCUSED_OPTION_INDEX"; payload: number};
+  | {type: "SET_FOCUSED_OPTION_INDEX"; payload: number}
+  | {type: "SET_OPTIONS"; options: (Option | null)[]};
```
edizcelik, line 53: why can an option be `null`?

src/select/Select.tsx
```diff
@@ -44,9 +45,19 @@ function SelectComponent<T extends Option = Option>(
     () => selectRef.current
   );
 
+  useEffect(() => {
+    if (props.options.length !== selectOwnState.options.length) {
```
edizcelik, line 49: This feels unsafe but I'm not sure if it worths doing a more proper equality check. @Hipo/frontend-team open to any suggestions here.

gulcinuras: As @Anlerkan suggested, I removed the `options` field from `selectOwnState` and directly set it in the context state from the props with `generateSelectState` function

Anlerkan (approved):

Anlerkan:
Let's update the PR title

PR #228, opened 2023-03-08
chore: Upgrade to react 18, and other deps

jamcry:
### Description
- Upgrade to React 18, and upgrade other dependencies
- Update peer dependencies to support both react 17 and 18

### Note
- Tested both with React 17 and 18 projects, seems to be working fine!


.storybook/webpack.config.js
```diff
@@ -34,8 +34,7 @@ module.exports = ({ config }) => {
             "@babel/plugin-proposal-nullish-coalescing-operator"
           ]
         }
-      },
-      require.resolve("react-docgen-typescript-loader")
```
jamcry, line 38: This package seems to be not maintained anymore (last change 3 years) and it was causing problems with typescript.
As far as I can see, it's for auto-generating documents from typescript comments, but I guess we don't need it at the moment. cc @mucahit

.gitignore
```diff
@@ -1,7 +1,6 @@
 .idea
 .vscode
 node_modules
-dist
```
jamcry, line 4: This is a temporary change to be able to test this branch on projects with commit hash

src/button/Button.tsx
```diff
@@ -62,7 +62,8 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
         autoFocus={shouldFocus}
         onClick={handleClick}
         disabled={isButtonDisabled}
-        {...rest}>
+        {...rest}
+      >
```
yasincaliskan, line 66: This change reason is a new prettier rule? 🤔 Wasn't the previous format better? What do you think?

jamcry: I've checked it again, I think they renamed a rule (`bracketSameLine`), that was the reason, fixed it now, thanks!

